PR #116, opened 2022-09-07
Fix types for **kwargs

The-Compiler:
`**kwargs` arguments should be typed according to their values (i.e. a *single* argument), not `kwargs` as a whole - see e.g. the [mypy docs](https://mypy.readthedocs.io/en/stable/getting_started.html?highlight=kwargs#more-function-signatures) (last snippet).

This fixes typing issues such as this one:

image

Weirdly enough, the Pydantic mypy plugin seems to hide this issue in this repository. If I comment out `plugins = pydantic.mypy` in `mypy.ini`, this is flagged in the tests and such... That seems like an issue in that plugin perhaps, but I have no idea of Pydantic, so I won't look at that in detail I'm afraid.

### DIFF
--- a/ical/event.py
+++ b/ical/event.py
@@ -233,7 +233,7 @@ class Event(ComponentModel):
     alarm: list[Alarm] = Field(alias="valarm", default_factory=list)
     """A grouping of reminder alarms for the event."""
 
-    def __init__(self, **data: dict[str, Any]) -> None:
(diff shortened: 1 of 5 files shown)


allenporter (approved):
Awesome thank you for catching this.

Curious, how'd you come across this? It's a fairly new repo with no dependencies on it yet :)

PR #117, opened 2022-09-07
Fix small doc issues

The-Compiler:
- `py.tests` should be `py.test`, but the dot isn't usually used anymore nowadays
- The quickstart example uses `e`, but only `event` is defined


allenporter (approved):

allenporter:
Thank you @The-Compiler !

PR #119, opened 2022-09-24
Simplify merged iterator based on PR feedback

allenporter:
Simplify merged iterator based on PR feedback by removing the peeking iterator and instead using the heap as the place to store the next item. This only reads what is strictly necessary from input iterators in `MergedIterator`

### DIFF
--- a/ical/iter.py
+++ b/ical/iter.py
@@ -71,31 +71,37 @@ class MergedIterator(Iterator[T]):
     def __init__(self, iters: list[Iterator[T]]):
         """Initialize MergedIterator."""
         self._iters = iters
-        self._heap: list[tuple[T, int]] = []
-        for iter_index, iterator in enumerate(iters):
-            try:
-                next_item = next(iterator)
-            except StopIteration:
-                # Iterator has been exhausted, don't add to heap
-                pass
-            else:
-                heapq.heappush(self._heap, (next_item, iter_index))
+        self._heap: list[tuple[T, int]] | None = None
 
     def __iter__(self) -> Iterator[T]:
         """Return this iterator."""
         return self
 
+    def _make_heap(self) -> None:
+        self._heap = []
+        for iter_index, iterator in enumerate(self._iters):
+            try:
+                next_item = next(iterator)
+            except StopIteration:
+                pass
+            else:
+                heapq.heappush(self._heap, (next_item, iter_index))
+
     def __next__(self) -> T:
         """Produce the next item from the merged set."""
+
+        if self._heap is None:
+            self._make_heap()
+
         if not self._heap:
             raise StopIteration()
+
         (item, iter_index) = heapq.heappop(self._heap)
         iterator = self._iters[iter_index]
         try:
             next_item = next(iterator)
         except StopIteration:
-            # Iterator will not be added back to the heap
-            pass
+            pass  # Iterator not added back to heap
         else:
             heapq.heappush(self._heap, (next_item, iter_index))
         return item
--- a/ical/iter.py
+++ b/ical/iter.py
@@ -65,42 +65,16 @@ class RecurIterable(Iterable[T]):
         return RecurIterator(self._item_cb, iter(self._recur))
 
 
-class PeekingIterator(Iterator[T]):
-    """An iterator with a preview of the next item.
-
-    The primary purpose is to implement a merged iterator where it is needed to
-    see the next item in the iterator in order to decide which child iterator
-    to pull from.
-    """
-
-    def __init__(self, iterator: Iterator[T]):
-        """Initialize PeekingIterator."""
-        self._iterator = iterator
-        self._next = next(self._iterator, None)
-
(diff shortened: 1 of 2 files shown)


ical/iter.py
```diff
@@ -65,58 +65,30 @@ def __iter__(self) -> Iterator[T]:
...
     """An iterator with a merged sorted view of the underlying sorted iterators."""
 
     def __init__(self, iters: list[Iterator[T]]):
         """Initialize MergedIterator."""
-        self._iters = [PeekingIterator(iterator) for iterator in iters]
+        self._heap: list[tuple[T, Iterator[T]]] = []
+        for iterator in iters:
+            next_item: T | None = next(iterator, None)
+            if next_item:
+                heapq.heappush(self._heap, (next_item, iterator))
```
make-github-pseudonymous-again, line 77: Maybe keep a separate copy of the list of iterators (with `tuple` maybe?), then enumerate the iterators and store the index of the iterator in the heap instead to break ties in a `heapq`-independent way.

ical/iter.py
```diff
@@ -65,58 +65,30 @@ def __iter__(self) -> Iterator[T]:
...
-            if peekd:
-                heapq.heappush(heap, (peekd, iterator))
-        if not heap:
+        if not self._heap:
             raise StopIteration()
-        (_, iterator) = heapq.heappop(heap)
-        return next(iterator)
+        (item, iterator) = heapq.heappop(self._heap)
+        next_item = next(iterator, None)
+        if next_item:
```
make-github-pseudonymous-again, line 89: Maybe it would be better to use `try`/`except` here, to be fully general. Iterators items are not necessarily truthy or not `None`.

make-github-pseudonymous-again (commented):
> This avoids rebuilding the heap each time and iterating through all recurring events.

This part would be better described as "only reading what is strictly necessary from input iterators in `MergedIterator`.

ical/iter.py
allenporter: I was writing a test to exercise the other issue you pointed (basically values in the list that would evaluate truthiness of values in the iterator) and it exercised the issue you're pointing out here too. Basically, the heap was trying to compare the iterators and failed:
```
           TypeError: '<' not supported between instances of 'list_iterator' and 'list_iterator'
```
I've updated this to have the indexes into a list of iterators as you suggested.

ical/iter.py
allenporter: Updated, and I added some tests to reproduce the issues you pointed out if this were to be used more generally. Agree, might as well make it correct since it looks general.

ical/iter.py
make-github-pseudonymous-again: Oh yes it actually crashes if you do not use indices! Good catch.

ical/iter.py
make-github-pseudonymous-again: Did you push the latest version?

allenporter: Whoops. Pushed.

make-github-pseudonymous-again (approved):
Perfect! The only thing that could have potential discussion is whether the heap should be initialized in `__init__` or in the first call to `__next__`. You could have `this._heap` set to `None` first, then to a `list` via `heapq`. Not sure what would be the most idiomatic way of handling a empty heap, maybe via `len`? This could also easily be implemented via a wrapper object.

The use case would be when lots of `MergedIterator` objects are created and discarded without even the first value being consumed.

make-github-pseudonymous-again (approved):
Perfect, again. I just realized there is also unnecessary iterator consumption before each yielded value since the next item is pushed onto the heap before the current is yielded. Handling this would make the implementation even more verbose so I am not sure it's worth it. But if you want to be more general... On the other hand an iterable implementation using `__iter__` would be more natural because intermediate state handling is implicit, for instance:

```py
def merge(iterators):
    index = list(iterators)
    heap = makeheap((x, i for i, x in enumerate(map(next, index))))
    while heap:
      (x, i) = heappop(heap)

      yield x

      try:
        y = next(index[i])
      except StopIteration:
        pass
      else:
        heappush(heap, (y, i))

allenporter:
Thanks for the review!

make-github-pseudonymous-again:
@allenporter Just realized [`heapq.merge`](https://docs.python.org/3/library/heapq.html#heapq.merge) exists! I vote for using it instead of our reimplementation. What do you think?

allenporter:
@make-github-pseudonymous-again wow, that is awesome! cool that it also supports the key function, etc. Seems like exactly what is needed here and can't believe we missed this before!

allenporter:
btw I had to fix some other issues similar to what you fixed here. Namely, the original implementation was using the item contents for the heap sort comparisons, which was a drag on performance. I updated this to instead just order all the timestamps instead and introduced a `SortableItem` to separate the key from the contents, and a benchmark in `test_benchmark_merged_iter` that showed the improvement.  I am optimistic that the key function they provide can make that unnecessary.

PR #123, opened 2022-10-21
Silence mypy errors for some existing problems

allenporter:
Silence mypy errors for some existing problems


allenporter:
These are cases where the default values / required field and typing are not currently matching and needs to be corrected. Adding as a silence for now.

PR #193, opened 2023-06-09
Fixed and tested misleading error messages on creating Alarm

mreiche:


tests/test_alarm.py
```diff
@@ -48,9 +48,11 @@ def test_duration_and_repeat() -> None:
 
 def test_display_required_fields() -> None:
     """Test required fields for action DISPLAY."""
-    with pytest.raises(ValidationError):
+    with pytest.raises(ValidationError) as e:
         Alarm(action="DISPLAY", trigger=datetime.timedelta(minutes=-5))
 
+    assert "Description value is required for action DISPLAY" in e.value.raw_errors[0].exc.args[0]
+
```
allenporter, line 55: I believe you can use the built in matcher for this:

```suggestion
    with pytest.raises(ValidationError, match="Description value is required for action DISPLAY"):
        Alarm(action="DISPLAY", trigger=datetime.timedelta(minutes=-5))
```

(here and below)

allenporter (changes requested):
Apologies, I missed this PR! Thank you for sending. I have one comment about test simplification.

tests/test_alarm.py
mreiche: Well, thats way better. Thanks for the hint.

allenporter (approved):

allenporter:
Thank you @mreiche 👍🏼

PR #215, opened 2023-08-20
Replace datetime.utcnow

cdce8p:
Starting with Python 3.12 `datetime.utcnow` will be deprecated.
https://docs.python.org/3.12/library/datetime.html#datetime.datetime.utcnow

### DIFF
--- a/ical/util.py
+++ b/ical/util.py
@@ -19,7 +19,7 @@ PRODID = "github.com/allenporter/ical"
 
 def dtstamp_factory() -> datetime.datetime:
     """Factory method for new event timestamps to facilitate mocking."""
-    return datetime.datetime.now(tz=datetime.timezone.utc).replace(tzinfo=None)
+    return datetime.datetime.now(tz=datetime.timezone.utc)
 
 
 def uid_factory() -> str:
--- a/ical/util.py
+++ b/ical/util.py
@@ -19,7 +19,7 @@ PRODID = "github.com/allenporter/ical"
 
 def dtstamp_factory() -> datetime.datetime:
     """Factory method for new event timestamps to facilitate mocking."""
-    return datetime.datetime.utcnow()
+    return datetime.datetime.now(tz=datetime.timezone.utc).replace(tzinfo=None)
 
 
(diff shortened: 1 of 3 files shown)


allenporter:
Similar to my comment on the home assistant PR, I think its more helpful to continue to use the deprecated method given the guidance `it is preferred to use aware datetimes to represent times in UTC` That is, meeting the spirit of the deprecation involves stopping using TZ aware times even for UTC.  I'd rather leave it to indicate there is more work to cleanup here. (This isn't scheduled for removal, right?)

cdce8p:
> Similar to my comment on the home assistant PR, I think its more helpful to continue to use the deprecated method given the guidance `it is preferred to use aware datetimes to represent times in UTC` That is, meeting the spirit of the deprecation involves stopping using TZ aware times even for UTC. I'd rather leave it to indicate there is more work to cleanup here.

🤔 I would say it depends. The main change left would be to remove `replace(tzinfo=None)` to use a timezone aware datetime object. However, I haven't tested if that would be a breaking change. Making the tests pass is easy https://github.com/allenporter/ical/pull/215/commits/7a1d1614bc73a521995b988fd62e1676e56fd320.
Overall I would say a strict replacement still has it uses. If for nothing else than to prevent it accidentally breaking with future Python versions.

In the end adding a `TODO` comment is almost as useful as seeing the deprecation warning. Just with the difference that there might be so many warnings that you miss the important ones when they show up.


> (This isn't scheduled for removal yet, right?)

Not yet, AFAIK

allenporter:
Yeah, i think its a larger change -- not necessarily expecting you to make it.  date-times in this library are a beast.

I think the existing use of this may result in unintended behavior -- a date without a timezone is assumed to be as floating in the local timezone. So this is actually a good thing to fix, since its not expected its actually utc.

cdce8p:
> Yeah, i think its a larger change -- not necessarily expecting you to make it. date-times in this library are a beast.

Thought that might be the case. I'm going to close this PR then.
No need to fix that right now. We can filter the warning in Home Assistant once we support Python 3.12.

PR #258, opened 2023-12-10
Add default end date/time for events with no end time or duration

### LINKED ISSUE #257: Event does not have correct default when no end or duration is specified
**Describe the bug**

Calendar events that do not specify `DTEND` or `DURATION` do not respect default in RFC 2445, and instead raise exception `ValueError("Unexpected state with no duration or dtend")`.

**To Reproduce**

Here's the contents of an ICS file I am trying to use.

```
BEGIN:VCALENDAR
PRODID:-//127.0.0.1//Waste Calendar Generator//
VERSION:2.0
X-WR-CALNAME:Bins Schedule
X-WR-CALDESC:Bins Schedule
X-WR-TIMEZONE:Europe/London
BEGIN:VEVENT
DTSTART;VALUE=DATE:20231212
SUMMARY:Black Bin Collection
END:VEVENT
BEGIN:VEVENT
DTSTART;VALUE=DATE:20231219
SUMMARY:Green Bin Collection
END:VEVENT
BEGIN:VEVENT
DTSTART;VALUE=DATE:20231219
SUMMARY:Blue Bin Collection
END:VEVENT
BEGIN:VEVENT
DTSTART;VALUE=DATE:20231228
SUMMARY:Black Bin Collection
END:VEVENT
END:VCALENDAR
```

**Expected behavior**

The library should treat infer a sensible default where a duration or end date/time has not been specified.

According to RFC 2445, section 4.6.1:

>For cases where a "VEVENT" calendar component specifies a "DTSTART" property with a DATE data type but no "DTEND" property, the events non-inclusive end is the end of the calendar date specified by the "DTSTART" property. For cases where a "VEVENT" calendar component specifies a "DTSTART" property with a DATE-TIME data type but no "DTEND" property, the event ends on the same calendar date and time of day specified by the "DTSTART" property.

**Additional context**

I'm trying to use an ICS file provided by my local government authority that describes bin collections and integrate that with my Home Assistant. I do not have any control over the ICS file itself, but it looks to be valid (at least with respect to absent `DTEND`, `DURATION`).

I'm also working on a fix, and will open a PR for this.

dannycjones:
Hey! Looking to fix #257.

Before this change, some events would be parsed and result in a ValueError. According to RFC 5545 section 3.6.1, there are default behaviors to be assumed in these cases - namely that events with a starting date should assume the event is one full day, while events with a starting time and date should assume zero duration.

Please let me know if there's any style conventions or test conventions to fix, I'm no Python expert.


ical/event.py
```diff
@@ -282,11 +287,9 @@ def end_datetime(self) -> datetime.datetime:
     @property
     def computed_duration(self) -> datetime.timedelta:
         """Return the event duration."""
-        if self.end:
-            return self.end - self.start
-        if not self.duration:
-            raise ValueError("Invalid state, expected end or duration")
-        return self.duration
+        if self.duration != None:
```
allenporter, line 290:
```suggestion
        if self.duration is not None:
```
prefering an explicit none check rather than comparison

tests/test_event.py
```diff
@@ -176,6 +176,22 @@ def test_start_end_same_type() -> None:
...
 
 
+def test_no_end_time_or_dur() -> None:
+    """Verify that events with no end time or duration will use correct defaults."""
+
+    day_event = Event(
+        summary=SUMMARY,
+        dtstart=date(2022, 9, 9)
+    )
+    assert day_event.end == date(2022, 9, 10)
```
allenporter, line 186: Mind adding tests for duration and `computed_duration` while you're here?

allenporter (changes requested):
Looks great overall, thank you for the fix and I see where this was missed from the rfc.

Would you also be up for adding an ICS file in `tests/testdata` and verifying `test_calendar_stream.py`? You can refresh the snapshots with `py.test tests/test_calendar_stream.py --snapshot-update`.  The unit tests are great, though i also like building up the collection of ics files to test with as part of fixing the bugs, and this will also show that the behavior when serializing the content back out to disk (your change doesn't explicitly set an end field the ics content, which seems like the right behavior)

allenporter:
> Looks great overall, thank you for the fix and I see where this was missed from the rfc.
> 
> Would you also be up for adding an ICS file in `tests/testdata` and verifying `test_calendar_stream.py`? You can refresh the snapshots with `py.test tests/test_calendar_stream.py --snapshot-update`. The unit tests are great, though i also like building up the collection of ics files to test with as part of fixing the bugs, and this will also show that the behavior when serializing the content back out to disk (your change doesn't explicitly set an end field the ics content, which seems like the right behavior)

I gave you the wrong instructions here. I thought I had upgraded this library to use syrup snapshots, but i got it mixed up with another library. It's `--update-goldens`

allenporter:
> > Looks great overall, thank you for the fix and I see where this was missed from the rfc.
> > Would you also be up for adding an ICS file in `tests/testdata` and verifying `test_calendar_stream.py`? You can refresh the snapshots with `py.test tests/test_calendar_stream.py --snapshot-update`. The unit tests are great, though i also like building up the collection of ics files to test with as part of fixing the bugs, and this will also show that the behavior when serializing the content back out to disk (your change doesn't explicitly set an end field the ics content, which seems like the right behavior)
> 
> I gave you the wrong instructions here. I thought I had upgraded this library to use syrup snapshots, but i got it mixed up with another library. It's `--update-goldens`

Actually -- we already have an example test file that does not have a duration, however the existing tests don't catch this. In I  https://github.com/allenporter/ical/pull/259 i just added a test that coincidentally failed due to the error that you are fixing here, but i had to exclude it since your fix is not in yet. If you sync to head and remove the guard for `rrule-yearly.yaml` that I just added we should be good to go without you need to add another example ics file.

tests/test_event.py
dannycjones: Good idea! Added.

dannycjones:
Thanks for the quick feedback. Just pushed those changes.

allenporter:
Some of the tests are failing because ical is not preserving the extended attributes (not really supported yet) and some other sort orders change. You can either remove that yaml test (given we have another test covering some cases), remove the attributes, or add to the yaml (like some other tests have) where it has a different ordering for the serialized output.

dannycjones:
I'll drop them for now since we already have the other test. :)

allenporter (approved):
Looks great, thanks @dannycjones 👍🏼

PR #327, opened 2024-04-27
Improve type hinting of registered types

olejorgenb:
Pyright does not understand the types decorated by `Registery.register`. Eg.: `Event.attendees` is inferred as `Unknown`

Using a typevar represents accurately how the decorator affect the decorated type (ie. not at all)

Before:

image

After:

image


allenporter (approved):

allenporter:
Thank you @olejorgenb 👍🏼

PR #328, opened 2024-04-27
Fix typo and wrong value in default_factory for CalendarStream.calendars

olejorgenb:
Suppose it's exceedingly rare that a stream contain no calendars, but better safe than sorry


allenporter (approved):
Thank you!  My impression is this is very important as it otherwise means all instances share the same list.

allenporter:
(I realize you're saying in practice it was a typo and didn't even work, but agree fixing is way way better!)

allenporter:
Thank you @olejorgenb 👍🏼

PR #384, opened 2024-06-30
Move version to const

Thomas55555:
I've reverted this [PR](https://github.com/home-assistant/core/pull/120276) and applied this changes here. This prevents the blocking call in the event loop in Home Assistant.

### DIFF
--- a/ical/util.py
+++ b/ical/util.py
@@ -15,6 +15,7 @@ __all__ = [
 
 MIDNIGHT = datetime.time()
 PRODID = "github.com/allenporter/ical"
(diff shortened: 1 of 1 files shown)


allenporter (approved):
Looks good, thanks. Are you are sure there are not also timezone issues too?

Thomas55555:
I can't tell for sure, but at least I don't receive any warnings.

allenporter:
Maybe I can preload those too in #386. The complexity added to gcal sync to workaround has made tests flaky.

PR #453, opened 2025-01-03
Add an interface for more efficient rrule parsing

allenporter:


ical/parsing/property.py
```diff
@@ -88,6 +90,39 @@ def ics(self) -> str:
...
+
+        Will raise a ValueError on failure.
+        """
+        # Lines can be in either of these formats where the parameters are optional:
+        # RRULE:FREQ=WEEKLY;COUNT=10
+        # RDATE;VALUE=DATE:19970304T080000Z
+        # RDATE:19970304T080000Z
+        if contentline.find(":") == -1:
+            raise ValueError(f"Expected ':' in contentline: {contentline}")
+        name, value = contentline.split(":", 1)
```
elupus, line 109: Use partition instead of split. Now it searches the string twice.
name, sep, value = contentline.partition(":")
if sep is None:
  raise ..

allenporter: Fixed, thanks!

elupus: You forgot to remove the old calculation of name and value.

ical/parsing/property.py
```diff
@@ -88,6 +90,40 @@ def ics(self) -> str:
...
+        Will raise a ValueError on failure.
+        """
+        # Lines can be in either of these formats where the parameters are optional:
+        # RRULE:FREQ=WEEKLY;COUNT=10
+        # RDATE;VALUE=DATE:19970304T080000Z
+        # RDATE:19970304T080000Z
+        name, sep, value = contentline.partition(":")
+        if sep is None:
+            raise ValueError(f"Expected ':' in contentline: {contentline}")
+        name, value = contentline.split(":", 1)
```
elupus, line 110:
```suggestion
```

ical/parsing/property.py
allenporter: Yes, saw this, and fixing. also the sep is None check doesn't work. adding tests and coverage still.

elupus: Ah right it give empty string not None.

ical/parsing/component.py
```diff
@@ -127,3 +127,10 @@ def parse_content(content: str) -> list[ParsedComponent]:
 def encode_content(components: list[ParsedComponent]) -> str:
     """Encode a set of parsed properties into content."""
     return "\n".join([component.ics() for component in components])
+
+
+def unfolded_lines(content: str) -> Generator[str, None, None]:
+    """Read content and unfold lines."""
+    content = FOLD_RE.sub("", content)
+    for line in LINES_RE.split(content):
+        yield line
```
elupus, line 136: you can do `yield from LINES_RE.split(content)`

allenporter: Nice, thanks, updated.

ical/parsing/property.py
allenporter: OK this should be cleaned up. Want to take another look?

tests/parsing/__snapshots__/test_property.ambr
```diff
@@ -0,0 +1,99 @@
+# serializer version: 1
+# name: test_from_ics[attendee]
+  list([
+    ParsedProperty(name='begin', value='VEVENT', params=None),
+    ParsedProperty(name='attendee', value='mailto: jsmith@example.com', params=[ParsedPropertyParameter(name='RSVP', values=['TRUE']), ParsedPropertyParameter(name='ROLE', values=['REQ-PARTICIPANT'])]),
+    ParsedProperty(name='attendee', value='jdoe@example.com","mailto:jqpublic @example.com":mailto:jsmith@example.com', params=[ParsedPropertyParameter(name='DELEGATED-TO', values=['"mailto'])]),
+    ParsedProperty(name='end', value='VEVENT', params=None),
```
allenporter, line 7: By the way, this snapshot contains examples where this heuristic does not parse content correctly, like the grammar does. This is an example of the type of bug that most other icalendar libraries have when implementing the spec, which motivated using pyparsing with a real grammar.

The "bugs" are recorded here and can be spotted when running tests that compare both parsing approaches.

ical/parsing/property.py
allenporter: I'm going to merge but feel free to follow up with any additional feedback.

PR #464, opened 2025-03-30
Support for zero-length paramtext...

tmcqueen-materials:
....as allowed by RFC5545 grammer (bottom page 10, `paramtext = *SAFE-CHAR`).

Found in-the-wild with an Apple iCal ics file with `X-APPLE-FILENAME=` as param, without this PR an exception is thrown. Pyparser `Word` expressly does not match zero length strings.

I am not sure this is the best way to tackle this, but it fixes all test cases I am able to find.

### DIFF
--- a/ical/parsing/parser.py
+++ b/ical/parsing/parser.py
@@ -67,7 +67,7 @@ def _create_parser() -> ParserElement:
     param_name = Or([iana_token, x_name])
     # rfc5545 Q-SAFE-CHAR is any character except CONTROL and DQUOTE which is
     # close enough to the pyparsing provided parser element
-    param_text = Word(SAFE_CHAR)
+    param_text = Word(SAFE_CHAR) | ""
     quoted_string = QuotedString('"')
 
     param_value = Or([param_text, quoted_string])
--- a/ical/parsing/property.py
+++ b/ical/parsing/property.py
@@ -142,7 +142,7 @@ def parse_property_params(
         for pair in parsed_params[PARSE_PARAMS]:
             params.append(
                 ParsedPropertyParameter(
-                    name=pair[PARSE_PARAM_NAME], values=pair.get(PARSE_PARAM_VALUE, "")
+                    name=pair[PARSE_PARAM_NAME], values=pair.get(PARSE_PARAM_VALUE, [""])
                 )
             )
     return params
--- a/ical/parsing/property.py
+++ b/ical/parsing/property.py
@@ -142,7 +142,7 @@ def parse_property_params(
         for pair in parsed_params[PARSE_PARAMS]:
             params.append(
                 ParsedPropertyParameter(
-                    name=pair[PARSE_PARAM_NAME], values=pair[PARSE_PARAM_VALUE]
+                    name=pair[PARSE_PARAM_NAME], values=pair[PARSE_PARAM_VALUE] if PARSE_PARAM_VALUE in pair else [""]
                 )
             )
     return params
--- a/ical/parsing/property.py
+++ b/ical/parsing/property.py
@@ -142,7 +142,7 @@ def parse_property_params(
         for pair in parsed_params[PARSE_PARAMS]:
             params.append(
                 ParsedPropertyParameter(
-                    name=pair[PARSE_PARAM_NAME], values=pair[PARSE_PARAM_VALUE] if PARSE_PARAM_VALUE in pair else [""]
+                    name=pair[PARSE_PARAM_NAME], values=pair.get(PARSE_PARAM_VALUE, "")
                 )
             )
     return params
--- a/tests/test_calendar_stream.py
+++ b/tests/test_calendar_stream.py
@@ -133,7 +133,7 @@ def test_multiple_calendars() -> None:
             """))
 
 def test_blank_param_value() -> None:
-    ics = IcsCalendarStream.calendar_from_ics(
+    IcsCalendarStream.calendar_from_ics(
         textwrap.dedent("""\
             BEGIN:VCALENDAR
             PRODID:-//example//1.2.3
--- a/tests/test_calendar_stream.py
+++ b/tests/test_calendar_stream.py
@@ -131,3 +131,52 @@ def test_multiple_calendars() -> None:
                 VERSION:2.0
                 END:VCALENDAR
             """))
+
+def test_blank_param_value() -> None:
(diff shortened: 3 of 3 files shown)


ical/parsing/parser.py
```diff
@@ -67,7 +67,7 @@ def _create_parser() -> ParserElement:
     param_name = Or([iana_token, x_name])
     # rfc5545 Q-SAFE-CHAR is any character except CONTROL and DQUOTE which is
     # close enough to the pyparsing provided parser element
-    param_text = Word(SAFE_CHAR)
+    param_text = Word(SAFE_CHAR) | ""
```
allenporter, line 70: Is this equivalent to the `Or` syntax above? If so please use that so its consistent with the existing code.

allenporter (changes requested):
Thank you for this contribution. Can you also add tests that exercise this case? Also it looks like the quoted string may have the same problem so would also be nice to exercise that case as well.

Thank you!

ical/parsing/parser.py
tmcqueen-materials: @allenporter : It is not exactly equivalent, this is actually making it `Opt` (Optional). I could change it to `Opt(Word(SAFE_CHAR))` if you would prefer (and add the Opt import above).

allenporter: Thanks i see this in the doc. We're also using other expression operators above like "+" so this seems fine.

tmcqueen-materials:
@allenporter : Tests for both fully blank values, and empty quoted string values have been added. I have confirmed they fail without the fixes of this PR, and pass with.

Regarding quoted-string, no additional changes are needed as `QuotedString` already properly handles the case of nothing between the quotation marks.

ical/parsing/property.py
```diff
@@ -142,7 +142,7 @@ def parse_property_params(
         for pair in parsed_params[PARSE_PARAMS]:
             params.append(
                 ParsedPropertyParameter(
-                    name=pair[PARSE_PARAM_NAME], values=pair[PARSE_PARAM_VALUE]
+                    name=pair[PARSE_PARAM_NAME], values=pair[PARSE_PARAM_VALUE] if PARSE_PARAM_VALUE in pair else [""]
```
allenporter, line 145: I believe this is equivalent to this:

```suggestion
                    name=pair[PARSE_PARAM_NAME], values=pair.get(PARSE_PARAM_VALUE, "")
```

tests/test_calendar_stream.py
```diff
@@ -131,3 +131,42 @@ def test_multiple_calendars() -> None:
                 VERSION:2.0
                 END:VCALENDAR
             """))
+
+def test_blank_param_value() -> None:
+    ics = IcsCalendarStream.calendar_from_ics(
```
allenporter, line 136:
```suggestion
    IcsCalendarStream.calendar_from_ics(
```

tests/test_calendar_stream.py
```diff
@@ -131,3 +131,42 @@ def test_multiple_calendars() -> None:
...
+        textwrap.dedent("""\
+            BEGIN:VCALENDAR
+            PRODID:-//example//1.2.3
+            VERSION:2.0
+            X-TEST-BLANK;VALUE=URI;X-TEST-BLANK-PARAM=:VALUE
+            END:VCALENDAR
+        """))
+
+def test_blank_quoted_param_value() -> None:
+    ics = IcsCalendarStream.calendar_from_ics(
```
allenporter, line 146:
```suggestion
    IcsCalendarStream.calendar_from_ics(
```

allenporter (changes requested):
(1) The tests don't seem to pass (not sure it was from my edits, seemed to not pass before that)
(2) Since it doesn't pass anyway,  test_calendar_stream.py really isn't for testing things at this low level granularity.  Any of these would probably be preferred:
  (a) https://github.com/allenporter/ical/tree/main/tests/examples if we just want to make sure some calendar in the wild can be parsed
  (b) https://github.com/allenporter/ical/tree/main/tests/testdata if we want to test how it is parsed in more detail
  (c) https://github.com/allenporter/ical/blob/main/tests/parsing/test_property.py if we to get into the weeds on property value specifics

tmcqueen-materials:
@allenporter : I do not understand why the tests are failing here? Here is from my system:
```
pytest tests/test_calendar_stream.py 
======================================================== test session starts =========================================================
platform linux -- Python 3.12.3, pytest-7.4.4, pluggy-1.4.0
rootdir: ${HOME}/code/ical
configfile: pytest.ini
plugins: syrupy-4.6.1, xdist-3.4.0
collected 139 items                                                                                                                  

tests/test_calendar_stream.py ................................................................................................ [ 69%]
...........................................                                                                                    [100%]

------------------------------------------------------ snapshot report summary -------------------------------------------------------
66 snapshots passed.
======================================================== 139 passed in 1.29s =========================================================
```

Any help to understand why the test fails in CI would be appreciated. Thanks!

tmcqueen-materials:
@allenporter : Not adverse to adding a more complete test as well, but these do test the individual bits that are changed by this / implied by the ABNF format, so I'd like to keep them. From the traces above, you can see they pass for me locally with this PR (and not without)... so they should pass in CI too?

tmcqueen-materials:
@allenporter : Figured it out. The default return in `property.py` needs to be `[""]` and not `""`. The reason for this is that as constructed it expects an array of one or more values.

allenporter:
> @allenporter : Figured it out. The default return in `property.py` needs to be `[""]` and not `""`. The reason for this is that as constructed it expects an array of one or more values.

OK it was my patch sorry about that!

allenporter:
> @allenporter : Not adverse to adding a more complete test as well, but these do test the individual bits that are changed by this / implied by the ABNF format, so I'd like to keep them. From the traces above, you can see they pass for me locally with this PR (and not without)... so they should pass in CI too?

Oh sorry I wasn't saying the tests were incomplete or they should be removed, this is more of a style/structure thing.  This is not the spot in the test directory where this kind of detail is tested. If you don't want to move it, then I will do it.

tmcqueen-materials:
@allenporter : Guess it was a good test that the tests were working =).

Since you are more familiar with where exactly you want these tests, if you could move them, that would be great.

Thanks again for the nice project! I tried several ics python parsers and this one is definitely the nicest =).

PR #484, opened 2025-05-24
Parse without pyparsing

### LINKED ISSUE #469: Crashes with danish characters
**Describe the bug**
Google Calendar outputs a file containting the line

```
ATTACH;FILENAME=Fødselsdag_40.pdf:https://someurl.com
```

This gives the error

```
  File "/home/vscode/.local/lib/python3.12/site-packages/ical/parsing/parser.py", line 111, in parse_contentlines
    return [parser.parse_string(line, parse_all=True) for line in lines if line]
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vscode/.local/lib/python3.12/site-packages/pyparsing/core.py", line 1219, in parse_string
    raise exc.with_traceback(None)
pyparsing.exceptions.ParseException: Expected ':', found 'ødselsdag'  (at char 17), (line:1, col:18)

```

**To Reproduce**
Create an ics file with the line above in an event and parse it.

**Expected behavior**
'ø' (and other) should be treated as a normal word character.

**Additional context**
-

AThomsen:
I don't expect this to be merged as it. It's more to start a discussion :-)

Is it overkill to use pyparsing for this library? The ical format seems not that complicated, but I'm probably missing something.

I've spend some hours implementing this basic parser just for fun: 
- All tests passes (except one where it expects a pyparsing exception to be thrown).
- I believe it's faster than using pyparsing.
- It handles unicode, but it's probably not as strict as the original implementation with regards to allowed characters.
- It may need to handle certain escape characters (not sure if this happens elsewhere in the library.

Fixes #469 (which I badly need for Home Assistant, as my calendars are currently broken).

### DIFF
--- a/ical/parsing/component.py
+++ b/ical/parsing/component.py
@@ -79,7 +79,6 @@ def _fold(contentline: str) -> list[str]:
         break_on_hyphens=False,
     )
 
-
 def parse_content(content: str) -> list[ParsedComponent]:
     """Parse content into raw properties.
 
@@ -94,8 +93,7 @@ def parse_content(content: str) -> list[ParsedComponent]:
     token_results = parse_contentlines(lines)
 
     stack: list[ParsedComponent] = [ParsedComponent(name="stream")]
-    for result in token_results:
-        result_dict = result.as_dict()
+    for result_dict in token_results:
         if PARSE_NAME not in result_dict:
             raise ValueError(
                 f"Missing fields {PARSE_NAME} or {PARSE_VALUE} in {result_dict}"
@@ -109,7 +107,7 @@ def parse_content(content: str) -> list[ParsedComponent]:
             component = stack.pop()
             if value != component.name:
                 raise ValueError(
-                    f"Unexpected '{result}', expected {ATTR_END}:{component.name}"
+                    f"Unexpected '{result_dict}', expected {ATTR_END}:{component.name}"
                 )
             stack[-1].components.append(component)
         else:
@@ -122,6 +120,7 @@ def parse_content(content: str) -> list[ParsedComponent]:
                 property_dict[PARSE_PARAMS] = property_params
             stack[-1].properties.append(ParsedProperty(**property_dict))
     return stack[0].components
+
 
 
 def encode_content(components: list[ParsedComponent]) -> str:
--- a/ical/parsing/parser.py
+++ b/ical/parsing/parser.py
@@ -25,24 +25,8 @@ Note: This specific example may be a bit confusing because one of the property p
 """
 
 import logging
-import threading
-from functools import cache
-from typing import cast
 from typing import Iterable
 
-from pyparsing import (
-    Combine,
-    Group,
-    Or,
-    ParserElement,
-    ParseResults,
-    QuotedString,
-    Word,
-    ZeroOrMore,
-    alphanums,
-    alphas,
-    nums,
-)
 
 from .const import (
     PARSE_NAME,
@@ -51,61 +35,90 @@ from .const import (
     PARSE_PARAMS,
     PARSE_VALUE,
 )
-from .unicode import SAFE_CHAR, VALUE_CHAR
 
 _LOGGER = logging.getLogger(__name__)
 
 
-@cache
-def _create_parser() -> ParserElement:
-    """Create rfc5545 parser."""
-    iana_token = Word(alphanums + "-")
-    vendor_id = Word(alphanums)
-    x_name = Combine("X-" + ZeroOrMore(vendor_id + "-") + alphas + nums + "-")
-    name = Or([iana_token, x_name])
+def parse_line(line: str) -> dict:
+    """Parse a single line."""
+    
+    dict_result = {}
+    dict_result.setdefault(PARSE_PARAMS, [])
 
-    param_name = Or([iana_token, x_name])
-    # rfc5545 Q-SAFE-CHAR is any character except CONTROL and DQUOTE which is
-    # close enough to the pyparsing provided parser element
-    param_text = Word(SAFE_CHAR) | ""
-    quoted_string = QuotedString('"')
+    pos = 0
 
-    param_value = Or([param_text, quoted_string])
-    # There are multiple levels of property parameter grouping since there can
-    # either be repreated property parameters with the same name or property
-    # s parameters with repeated values. A two level structure is used to grab
-    # both, which is then flattened when consuming the result.
-    param = Group(
-        param_name.set_results_name(PARSE_PARAM_NAME)
-        + "="
-        + param_value.set_results_name(PARSE_PARAM_VALUE, list_all_matches=True)
-        + ZeroOrMore(
-            "," + param_value.set_results_name(PARSE_PARAM_VALUE, list_all_matches=True)
-        )
-    ).set_results_name(PARSE_PARAMS, list_all_matches=True)
+    # parse NAME
+    while True:
+        assert pos < len(line), "Unexpected end of line"
+        char = line[pos]
+        if char == ';' or char == ':':
+            dict_result[PARSE_NAME] = line[0:pos]
+            break
+        pos += 1
 
-    contentline = (
-        name.set_results_name(PARSE_NAME)
-        + Group(ZeroOrMore(";" + param)).set_results_name(
-            PARSE_PARAMS, list_all_matches=True
-        )
-        + ":"
-        + Word(VALUE_CHAR)[0, 1].set_results_name(PARSE_VALUE)
-    )
-    contentline.set_whitespace_chars("")
-    if _LOGGER.isEnabledFor(logging.DEBUG):
-        contentline.set_debug(flag=True)
-    return cast(ParserElement, contentline)
+    # parse PARAMS if any
+    if line[pos] == ';':
+        pos += 1
+        params_start = pos
+        value_start = 0
+        while pos < len(line):
+            char = line[pos]
+            if char == '=':
+                # param name reached
+                param_name = line[params_start:pos]
+                pos += 1
+                
+                # Now read values. (list separated by comma)
+                param_values = []
+                all_values_read = False
+                while not all_values_read:
+                    value_start = pos
+                    quoted = False
+
+                    if line[pos] == '"':
+                        # read all in quotes
+                        quoted = True
+                        pos += 1
+
+                    value_read = False
+                    while not value_read:
+                        assert pos < len(line), "Unexpected end of line"
+                        char = line[pos]
+                        
+                        if (quoted and char == '"') or (not quoted and (char == ',' or char == ';' or char == ':')):
+                            if quoted:
+                                param_value = line[value_start + 1:pos]
+                                pos += 1
+                                char = line[pos]
+                            else:
+                                param_value = line[value_start:pos]
+
+                            param_values.append(param_value)
+                            value_read = True
+                            all_values_read = char != ','
+                        pos += 1
+                    
+                dict_result[PARSE_PARAMS].append(
+                    {PARSE_PARAM_NAME: param_name, PARSE_PARAM_VALUE: param_values}
+                )
+                
+                if char == ':':
+                    break
+                params_start = pos
+            pos += 1
+    else:
+        assert line[pos] == ':', "Expected ':' after property name"
+        pos += 1
+
+    value = line[pos:]
+    dict_result[PARSE_VALUE] = value
+    return dict_result        
 
 
-_parser_lock = threading.Lock()
 
-
-def parse_contentlines(lines: Iterable[str]) -> list[ParseResults]:
+def parse_contentlines(lines: Iterable[str]) -> list[dict]:
     """Parse a set of unfolded lines into parse results.
 
     Note, this method is not threadsafe and may be called from only one method at a time.
     """
-    with _parser_lock:
-        parser = _create_parser()
-        return [parser.parse_string(line, parse_all=True) for line in lines if line]
+    return [parse_line(line) for line in lines if line]
--- a/ical/parsing/property.py
+++ b/ical/parsing/property.py
@@ -128,7 +128,6 @@ class ParsedProperty:
             params=parsed_property_parameters or None,
         )
 
-
 def parse_property_params(
     parse_result_dict: dict[str, str | list]
 ) -> list[ParsedPropertyParameter]:
@@ -137,16 +136,12 @@ def parse_property_params(
         return []
     params: list[ParsedPropertyParameter] = []
     for parsed_params in parse_result_dict[PARSE_PARAMS]:
-        if not isinstance(parsed_params, dict) or PARSE_PARAMS not in parsed_params:
-            continue
-        for pair in parsed_params[PARSE_PARAMS]:
-            params.append(
-                ParsedPropertyParameter(
-                    name=pair[PARSE_PARAM_NAME], values=pair.get(PARSE_PARAM_VALUE, [""])
-                )
+        params.append(
+            ParsedPropertyParameter(
+                name=parsed_params[PARSE_PARAM_NAME], values=parsed_params.get(PARSE_PARAM_VALUE, [""])
             )
+        )
     return params
-
 
 def parse_basic_ics_properties(
     contentlines: Iterable[str],


allenporter (changes requested):
Thank you, appreciate the contribution, and I am definitely open to this. Pyparsing was used because the RFC specifies a grammar, so I thought it was useful to be fully conformant. I've see buggy ad-hoc implementations so it was nice to know that this is 100% correct.

Totally agree however that its overkill. I think it would be nice to keep both versions and continue running the unit tests both ways to verify they are the same. We can make an exception for the bug fix, of course.

I would like to add some additional unit tests for cases I may worry are missed by this implementation -- but otherwise agree, we want the tests should be fully exercising the spec, and if they're passing that should be sufficient. Let me double check for gaps.

allenporter (changes requested):
The parsing code has benchmarks. You can run the tests with `--benchmark-only` to show before and after improvements and add them to the PR description.

allenporter:
I'm adding https://github.com/allenporter/ical/pull/485 with some additional test cases. It captures two bugs (1) the original bug you reported and (2) an existing bug in parameter quoting, so we'll want to capture those fixes with this PR.

AThomsen:
Thanks for replying and looking into this!

I applied your comments and rebased my branch on your updates.

Also added some optimizations to the parser - as much as I could without breaking the snapshot comparison in the tests. 

Benchmark on my machine looks like this:
~~~
This implementation: `21 passed, 1176 skipped in 22.69s`
Current implementation (main): `21 passed, 1176 skipped in 24.05s`
~~~
Gains could be higher in a real-world scenario, as the new parser is thread-safe with no need for locking.

The following unit tests currently FAIL:
~~~
==================================================================== short test summary info ====================================================================
FAILED tests/parsing/test_component.py::test_invalid_contentlines[invalid_x-name-2] - Failed: DID NOT RAISE <class 'ical.exceptions.CalendarParseError'>
FAILED tests/parsing/test_component.py::test_invalid_contentlines[invalid_x-name-1] - Failed: DID NOT RAISE <class 'ical.exceptions.CalendarParseError'>
FAILED tests/parsing/test_component.py::test_invalid_contentlines[invalid_iana-token] - Failed: DID NOT RAISE <class 'ical.exceptions.CalendarParseError'>
FAILED tests/parsing/test_component.py::test_invalid_contentlines[invalid_param_name-1] - Failed: DID NOT RAISE <class 'ical.exceptions.CalendarParseError'>
FAILED tests/parsing/test_component.py::test_invalid_contentlines[invalid_param_name-2] - Failed: DID NOT RAISE <class 'ical.exceptions.CalendarParseError'>
FAILED tests/test_calendar_stream.py::test_invalid_ics[control-char-value] - Failed: DID NOT RAISE <class 'ical.exceptions.CalendarParseError'>
FAILED tests/test_calendar_stream.py::test_invalid_ics[control-char-name] - Failed: DID NOT RAISE <class 'ical.exceptions.CalendarParseError'>
FAILED tests/test_calendar_stream.py::test_invalid_ics[non-us-ascii-unicode] - Failed: DID NOT RAISE <class 'ical.exceptions.CalendarParseError'>
========================================================== 8 failed, 1168 passed, 21 skipped in 9.09s ==============================================
~~~
Besides the last one (where it's actually the test that's wrong), all the failing tests are about checking if the parsed data only uses characters according to spec. It's might be hard to implement this correctly (which is why I made this PR in the first place). 

While *writers* of ical data should totally conform to spec, is it really necessary  for *readers* to validate 100%?

allenporter:
"This implementation: `21 passed, 1176 skipped in 22.69s`" -- maybe double check that the skipped files are not adding a lto of overhead and only run the benchmark tests.

ical/parsing/parser.py
```diff
@@ -51,61 +35,90 @@
...
+                                char = line[pos]
+                            else:
+                                param_value = line[value_start:pos]
+
+                            param_values.append(param_value)
+                            value_read = True
+                            all_values_read = char != ','
+                        pos += 1
+                    
+                dict_result[PARSE_PARAMS].append(
```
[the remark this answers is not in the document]
allenporter: Also apply to other members of `dict_result` -- essentially drop that as a local variable.

ical/parsing/component.py
```diff
@@ -79,7 +79,6 @@ def _fold(contentline: str) -> list[str]:
         break_on_hyphens=False,
     )
 
-
```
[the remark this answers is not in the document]
allenporter: This isn't done, still other areas not passing lnt.

ical/parsing/parser.py
```diff
@@ -54,67 +39,127 @@
...
-        )
-    ).set_results_name(PARSE_PARAMS, list_all_matches=True)
+_RE_CONTROL_CHARS = re.compile("[\x00-\x08\x0A-\x1F\x7F]")
+_RE_NAME = re.compile("[A-Z0-9\-]+")
 
-    contentline = (
-        name.set_results_name(PARSE_NAME)
-        + Group(ZeroOrMore(";" + param)).set_results_name(
-            PARSE_PARAMS, list_all_matches=True
+@cache
```
allenporter, line 49: Curious what kind of performance benefits we see vs memory usage of caching this. Maybe an LRU with fixed size could be more appropriate here.

allenporter (commented):
You can use something like `pytest tests/parsing/test_component.py --benchmark-only` to view detailed benchmark results of just these tests

allenporter (commented):
While pyparing is a complex library, it seems simple to specify the grammar compared handrolliing these many loops. Unless the code can be further simplified (so far its adding 20+ lines)

If we don't see significant performance wins, then we can explore using pyparsing with a loosed character set then use the same regular expressions to validate parameter names outside of pyparsing.

Thanks for your work to improve this.

allenporter:
I may have found a way to do this with pyparsing here https://github.com/allenporter/ical/pull/487

ical/parsing/parser.py
```diff
@@ -54,67 +37,104 @@
...
+    
+    dict_result = {}
+    pos = 0
+
+    # parse NAME
+    while True:
+        if pos >= len(line):
+            raise CalendarParseError(f"Unexpected end of line. Expected ';' or ':'", detailed_error = line)
+        char = line[pos]
+        if char == ';' or char == ':':
```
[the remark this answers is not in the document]
AThomsen: Oh, sorry. Python isn't my first language 😅

ical/parsing/parser.py
AThomsen: I'll re-add the caching decorator

allenporter: I'm worried about the memory usage of doing this. Whats the performance with and without? What does it look like with an LRU?

ical/parsing/parser.py
```diff
@@ -25,96 +25,133 @@
...
-    param = Group(
-        param_name.set_results_name(PARSE_PARAM_NAME)
-        + "="
-        + param_value.set_results_name(PARSE_PARAM_VALUE, list_all_matches=True)
-        + ZeroOrMore(
-            "," + param_value.set_results_name(PARSE_PARAM_VALUE, list_all_matches=True)
+def parse_line(line: str) -> ParsedProperty:
+    """Parse a single line."""
+    
+    result = ParsedProperty(name="", value="", params=None)
```
allenporter, line 43: Let's keep the ParsedProperty immutable and construct it at the end as the return value, similar to what I was saying about the dict return result.

ical/parsing/parser.py
```diff
@@ -25,96 +25,133 @@
...
+    line_len = len(line)
+    pos = 0
+
+    # parse NAME
+    while True:
+        if pos >= line_len:
+            raise CalendarParseError(f"Unexpected end of line. Expected ';' or ':'", detailed_error=line)
+        if line[pos] in (';', ':'):
```
allenporter, line 51: Do we loose any performance if we just get the index of ';' or ':' rather than looping? The looping here and below makes this more complex to follow so wondering if we actually see large benefits.

For example when searching through the parameter list we're doing 3 things at once needing to track the "all params ready" stuff.

Could we instead (1) search for the end of the param and (2) split by ","?

ical/parsing/parser.py
```diff
@@ -25,96 +25,133 @@
...
+                        param_value = line[param_value_start + 1:pos]
+                        param_value_read = True
+                        pos += 1
+                        char = line[pos]
+                    elif not quoted and char in (',', ';', ':'):
+                        param_value = line[param_value_start:pos]
+                        param_value_read = True
+
+                    if param_value_read:
+                        if not char in (',', ';', ':'):
```
allenporter, line 117: I think this only applies when exiting the quoting so move this inside that branch?

ical/parsing/parser.py
```diff
@@ -25,96 +25,133 @@
...
+            all_values_read = False
+            
+            while not all_values_read:
+                param_value_start = pos
+                quoted = False
+
+                if line[pos] == '"':
+                    # parameter value is quoted
+                    quoted = True
+                    pos += 1
```
allenporter, line 93: We can do something like this:

```suggestion
                if quoted := (line[pos] == '"'):
                    pos += 1
```

ical/parsing/parser.py
```diff
@@ -25,96 +25,133 @@
...
+
+                param_value_read = False
```
allenporter, line 95: Lets move param_value_start down here and already bake in the +1 for quoting and remove the +1 below

ical/parsing/parser.py
```diff
@@ -25,96 +25,133 @@
...
+                while not param_value_read:
+                    if pos >= line_len:
+                        if quoted:
+                            raise CalendarParseError(f"Unexpected end of line. Expected end of qouted string", detailed_error=line)
+                        else:
+                            raise CalendarParseError(f"Unexpected end of line. Expected ',', ';' or ':'", detailed_error=line)
+                    
+                    char = line[pos]
+
+                    if char == '"':
```
allenporter, line 105: Can you add an additional test cases in the quoted params file with something like this?

```
NAME;PARAM-NAME="":VALUE
NAME;PARAM-NAME="",Value2,"",VALUE4,"VALUE5":VALUE
```

This catches a bug in existing pyparsing quoted string since it expects more than one character, but the rfc allows it to be empty.

allenporter (changes requested):
Yes, now that th ebug is fixed, i'm still interested in the performance improvement. Thanks for running the benchmark, that is quite a large improvement.

I have a couple more comments, but I think this is close and we can improve in further rounds after merging.

ical/parsing/parser.py
```diff
@@ -25,96 +25,133 @@
...
+                
+            result.params.append(ParsedPropertyParameter(name=param_name, values=param_values))
+            
+            if not all_params_read:
+                # reset for next parameter
+                params_start = pos
+                pos += 1
+    else:
+        if line[pos] != ':':
+            raise CalendarParseError(f"Expected ':' after property name", detailed_error=line)
```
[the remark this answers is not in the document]
AThomsen: No, it's not hit :-)

AThomsen:
> > There are probably still some linting errors. I've yet to figure out how to run that locally.
> 
> I've added instructions here https://github.com/allenporter/ical/blob/main/CONTRIBUTING.md#running-pre-commit

Hmm...
```

❯ pre-commit
[INFO] Initializing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Initializing environment for https://github.com/psf/black.
[INFO] Initializing environment for https://github.com/charliermarsh/ruff-pre-commit.
[INFO] Initializing environment for https://github.com/codespell-project/codespell.
[INFO] Initializing environment for https://github.com/adrienverge/yamllint.git.
[INFO] Initializing environment for https://github.com/asottile/setup-cfg-fmt.
[INFO] Installing environment for https://github.com/pre-commit/pre-commit-hooks.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/psf/black.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/charliermarsh/ruff-pre-commit.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/codespell-project/codespell.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/adrienverge/yamllint.git.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/asottile/setup-cfg-fmt.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
trim trailing whitespace.............................(no files to check)Skipped
fix end of files.....................................(no files to check)Skipped
check yaml...........................................(no files to check)Skipped
check for added large files..........................(no files to check)Skipped
black................................................(no files to check)Skipped
ruff (legacy alias)..................................(no files to check)Skipped
mypy.................................................(no files to check)Skipped
codespell............................................(no files to check)Skipped
yamllint.............................................(no files to check)Skipped
setup-cfg-fmt........................................(no files to check)Skipped
```

AThomsen:
I followed the CONTRIBUTING guide (been using devcontainers for now), and it's now quote working as expected:
~~~
uv venv --python=3.13
Using CPython 3.13.3 interpreter at: /usr/bin/python3.13
Creating virtual environment at: .venv
Activate with: source .venv/bin/activate
~~~

This creates `.env/`, not `env/` so I had to change the `source` line to include the new "dot" path.

`pre-commit` does nothing (no files to check - see previous comment). Not sure if it's due to the env-path change.

I made some code changes as you suggested and added some tests. 

Attached are some benchmark comparisons:
* [orig-vs-new-w-caching.txt](https://github.com/user-attachments/files/20465125/orig-vs-new-w-caching.txt) - the compares the main branch (original implementation) to the new, with caching enabled.
* [new-cache-vs-nocache.txt](https://github.com/user-attachments/files/20465122/new-cache-vs-nocache.txt) - new implementation with caching vs same with no caching.

Based on these, I suggest we turn off caching. For a long-running server app like Home Assistant won't caching effectively mean that all calendar lines are kept in-memory "forever"?

As a last note, I would also suggest that you run the benchmarks yourself. I get quite varied timings on my system between test runs.

allenporter:
> > > There are probably still some linting errors. I've yet to figure out how to run that locally.
> > 
> > 
> > I've added instructions here https://github.com/allenporter/ical/blob/main/CONTRIBUTING.md#running-pre-commit
> 
> Hmm...
> 
> ```
> 
> ❯ pre-commit
> [INFO] Initializing environment for https://github.com/pre-commit/pre-commit-hooks.
> [INFO] Initializing environment for https://github.com/psf/black.
> [INFO] Initializing environment for https://github.com/charliermarsh/ruff-pre-commit.
> [INFO] Initializing environment for https://github.com/codespell-project/codespell.
> [INFO] Initializing environment for https://github.com/adrienverge/yamllint.git.
> [INFO] Initializing environment for https://github.com/asottile/setup-cfg-fmt.
> [INFO] Installing environment for https://github.com/pre-commit/pre-commit-hooks.
> [INFO] Once installed this environment will be reused.
> [INFO] This may take a few minutes...
> [INFO] Installing environment for https://github.com/psf/black.
> [INFO] Once installed this environment will be reused.
> [INFO] This may take a few minutes...
> [INFO] Installing environment for https://github.com/charliermarsh/ruff-pre-commit.
> [INFO] Once installed this environment will be reused.
> [INFO] This may take a few minutes...
> [INFO] Installing environment for https://github.com/codespell-project/codespell.
> [INFO] Once installed this environment will be reused.
> [INFO] This may take a few minutes...
> [INFO] Installing environment for https://github.com/adrienverge/yamllint.git.
> [INFO] Once installed this environment will be reused.
> [INFO] This may take a few minutes...
> [INFO] Installing environment for https://github.com/asottile/setup-cfg-fmt.
> [INFO] Once installed this environment will be reused.
> [INFO] This may take a few minutes...
> trim trailing whitespace.............................(no files to check)Skipped
> fix end of files.....................................(no files to check)Skipped
> check yaml...........................................(no files to check)Skipped
> check for added large files..........................(no files to check)Skipped
> black................................................(no files to check)Skipped
> ruff (legacy alias)..................................(no files to check)Skipped
> mypy.................................................(no files to check)Skipped
> codespell............................................(no files to check)Skipped
> yamllint.............................................(no files to check)Skipped
> setup-cfg-fmt........................................(no files to check)Skipped
> ```

OK this is likely because the files are committed. I think you can:
(1) add some blank lines to the changed files
(2) git add <filename>
(3) re-run pre-commit

allenporter:
> I followed the CONTRIBUTING guide (been using devcontainers for now), and it's now quote working as expected:
> 
> ```
> uv venv --python=3.13
> Using CPython 3.13.3 interpreter at: /usr/bin/python3.13
> Creating virtual environment at: .venv
> Activate with: source .venv/bin/activate
> ```
> 
> This creates `.env/`, not `env/` so I had to change the `source` line to include the new "dot" path.
> 

Thank you, fixed the docs which missed updating this when moving to `uv`

> `pre-commit` does nothing (no files to check - see previous comment). Not sure if it's due to the env-path change.
> 
> I made some code changes as you suggested and added some tests.
> 
> Attached are some benchmark comparisons:
> 
> * [orig-vs-new-w-caching.txt](https://github.com/user-attachments/files/20465125/orig-vs-new-w-caching.txt) - the compares the main branch (original implementation) to the new, with caching enabled.
> * [new-cache-vs-nocache.txt](https://github.com/user-attachments/files/20465122/new-cache-vs-nocache.txt) - new implementation with caching vs same with no caching.
> 
> Based on these, I suggest we turn off caching. For a long-running server app like Home Assistant won't caching effectively mean that all calendar lines are kept in-memory "forever"?

Thanks, appreciate this.

> 
> As a last note, I would also suggest that you run the benchmarks yourself. I get quite varied timings on my system between test runs.

I did run these and saw anywhere from a 3x and 10x speed up -- it all seemed quite significant.

allenporter:
> Btw: do you know if it's possible to have a quote sign (") inside a parameter value? And what would that look like?
> 
> "Friggin \"lasers\"!" ?

I think this part of the spec is relevant:

>    Property parameter values that contain the COLON, SEMICOLON, or COMMA
>   character separators MUST be specified as quoted-string text values.
>   Property parameter values MUST NOT contain the DQUOTE character.  The
>   DQUOTE character is used as a delimiter for parameter values that
>   contain restricted characters or URI text.  For example:

>     DESCRIPTION;ALTREP="cid:part1.0001@example.org":The Fall'98 Wild
>       Wizards Conference - - Las Vegas\, NV\, USA


It looks like the purpose of the quotes is strictly to handle  COLON, SEMICOLON, or COMMA encoding only and can't be used for strings that contain quotes.

tests/parsing/__snapshots__/test_property.ambr
```diff
@@ -48,6 +48,14 @@
     ParsedProperty(name='end', value='VCALENDAR', params=None),
   ])
 # ---
+# name: test_from_ics[params_empty]
+  list([
+    ParsedProperty(name='begin', value='VCALENDAR', params=None),
+    ParsedProperty(name='name', value='VALUE', params=[ParsedPropertyParameter(name='PARAM-NAME', values=['""'])]),
+    ParsedProperty(name='name', value='VALUE', params=[ParsedPropertyParameter(name='PARAM-NAME', values=['"",Value2,"",VALUE4,"VALUE5"'])]),
```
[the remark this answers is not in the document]
AThomsen: Well, this is from the `parse_basic_ics_properties` which I'm not sure why it's even there.

tests/parsing/__snapshots__/test_property.ambr
```diff
@@ -48,6 +48,14 @@
     ParsedProperty(name='end', value='VCALENDAR', params=None),
   ])
 # ---
+# name: test_from_ics[params_empty]
+  list([
+    ParsedProperty(name='begin', value='VCALENDAR', params=None),
+    ParsedProperty(name='name', value='VALUE', params=[ParsedPropertyParameter(name='PARAM-NAME', values=['""'])]),
```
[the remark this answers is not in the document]
AThomsen: Same with this one.

AThomsen:
Ok - pre-commit passes now :-)

I'll be gone for a few days, btw. Back sunday :-)

allenporter (approved):
Thank you very much for the contribution!

ical/parsing/parser.py
```diff
@@ -24,97 +24,153 @@
...
-        parser = _create_parser()
-        try:
-            return [parser.parse_string(line, parse_all=True) for line in lines if line]
-        except ParseException as err:
-            raise CalendarParseError(
-                f"Failed to parse calendar contents", detailed_error=str(err)
-            ) from err
+    try:
+        return [parse_line(line) for line in lines if line]
+    except Exception as err:
```
allenporter, line 173: I didn't realize this was catching `Exception`.  This is catching an `IndexError` for one of the tests I added because it fails the check `if (char := line[pos]) == '"':` 

I think we're going to need to restructure this code to simplify this, taking a look.

Issue #491, opened 2025-05-31
Confusing error message when TZID not recognized


jackschmidt:
**Describe the bug**
When the parser correctly finds a TZID error in a date-time, it also complains that the date-time does not match the pattern for a date.

**To Reproduce**
Here is a small reproducer, notice the typo in New Mork:
```
from ical.calendar_stream import IcsCalendarStream

IcsCalendarStream.calendar_from_ics("""
BEGIN:VCALENDAR
BEGIN:VEVENT
DTSTART;TZID=America/New_Mork:20250531T063000
DTEND;TZID=America/New_Mork:20250531T073000
SUMMARY:breakfast
END:VEVENT
END:VCALENDAR
... """)
```

produces

```
ical.exceptions.CalendarParseError: Failed to parse calendar EVENT component: Failed to validate: 20250531T063000, errors: (["Expected DATE-TIME TZID value 'America/New_Mork' to be valid timezone", "Expected value to match DATE pattern: '20250531T063000'"])
```

The first error reported is valid, but the second `"Expected value to match DATE pattern: '20250531T063000'"` is not reasonable.

**Expected behavior**

Just one error message:

```
ical.exceptions.CalendarParseError: Failed to parse calendar EVENT component: Failed to validate: 20250531T063000, errors: (["Expected DATE-TIME TZID value 'America/New_Mork' to be valid timezone"])
```

**Additional context**

This is from the bug report: https://github.com/home-assistant/core/issues/145493

We looked through the code, but couldn't find how two errors are reported, so don't have a fix.

The original bug report is probably POSIX vs Windows time zone identifiers. #468 

If we made a pull request to support those Windows timezones, would it be welcome?

allenporter:
Thank you for the excellent bug report!  Agree this is a valid issue and i've seen it before but did not previously prioritize it. Now in home assistant we serve these errors directly to the user so it is important to have them be friendly.

 I thought I could take a moment to explain what is going on here, if nothing else then is a note to future self/readers to fix.

## Background

The reason this happens is because these date fields are parsed either as a datetime or a date. The first error message is from attempting to parse it as a date-time. The second error is from attempting to parse it as a date.  What happens is the internal parser (a pydantic validator) will try one then the other, and it doesn't currently have enough information to differentiate these use cases.

## Detail: How the code works

The individual date-time and date parsing logic are in these places:
[...]

PR #498, opened 2025-06-01
Fix confusing message when TZID not recognized -- Minimal version

jackschmidt:
This is a minimal change to clarify the error message when a value cannot be validated as a Union[datetime, date]. Two error messages are displayed, but they more clearly indicate that one error message is for datetime, and one is for date.

We also made the fuller version you suggested in #491. I think the main use is for datetime+date, but this one is very simple and general.

### DIFF
--- a/ical/component.py
+++ b/ical/component.py
@@ -228,7 +228,9 @@ class ComponentModel(BaseModel):
                 )
                 errors.append(str(err))
                 continue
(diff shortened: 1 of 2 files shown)


allenporter:
I realize this can still be useful even with the other version.  Want to update the tests now since i merged the other version?

PR #501, opened 2025-06-05
Fix case-insensitive property names

jackschmidt:
According to [RFC2425 sec 5.8.2](https://datatracker.ietf.org/doc/html/rfc2425#section-5.8.2) and [RFC5545 sec 2](https://datatracker.ietf.org/doc/html/rfc5545#section-2) property names are case-insensitive. A football match calendar included a mixed case property name which caused an unexpected error, which this commit fixes.

This will fix home-assistant/core#146173 once the new version of ical is included there.

Limitations (due to time): I did not create a test to verify this behavior, and I didn't expand the upper-casing to other aspects. Parameter names and many values are also supposed to be case insensitive, but I did not have time to analyze the full scope and did not want to risk an overly broad fix.

### DIFF
--- a/ical/parsing/property.py
+++ b/ical/parsing/property.py
@@ -222,6 +222,7 @@ def _parse_line(line: str) -> ParsedProperty:
             if delimiter == ":":
                 break  # We are done with all parameters.
 
+    property_name = property_name.upper()
     if not _RE_NAME.fullmatch(property_name):
         raise CalendarParseError(
             f"Invalid property name '{property_name}'", detailed_error=line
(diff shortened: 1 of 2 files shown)


allenporter (changes requested):
Can you please add a test that reproduces the issue first?

jackschmidt:
I added a small test for case insensitive property names.

Why I had hesitated: I tried to find a nice example from the RFC, but it consistently uses upper case. There is a full ics file in the home-assistent/core#146173 issue, but I wasn't sure (a) how to use syrupy, and (b) if it is ok to take the "commercial" ics file and commit it (should I delete the description?). So probably good enough, I had just hoped to do better :smile:

allenporter (approved):

PR #502, opened 2025-06-09
Update tzif parsing to drop pyparsing dependency

AThomsen:
This patch replaces TZ parsing with a regex based parser. This means `pyparsing` is now no longer a dependency.

Performance-wise this is an improvement. Caching of available timezones (`for key in available_timezones(): ...` in `timezoneinfo.py`) is reduced from 1.5 sec to 0.4 sec on my machine.


AThomsen:
I'm not sure if the "validators" are still needed?

ical/tzif/tz_rule.py
```diff
@@ -200,64 +181,98 @@ def default_dst_offset(cls, values: dict[str, Any]) -> dict[str, Any]:
...
+    """Parser for TZ strings into Rule objects."""
 
-    hour = Combine(Opt(Word("+-")) + Word(nums))
-    tz_time = hour.set_results_name("hour") + Opt(
-        ":"
-        + Word(nums).set_results_name("minutes")
-        + Opt(":" + Word(nums).set_results_name("seconds"))
-    )
+    _str: str
+    _pos: int
```
allenporter, line 188:
```suggestion
```

Not needed?

ical/tzif/tz_rule.py
```diff
@@ -200,64 +181,98 @@ def default_dst_offset(cls, values: dict[str, Any]) -> dict[str, Any]:
...
-    is_julian_date = ",J" in tz_str
-    tz_rule = _create_parser(start_gator, is_julian_date)
-    try:
-        result = tz_rule.parse_string(tz_str, parse_all=True)
-    except ParseException as err:
-        raise ValueError(f"Unable to parse TZ string: {tz_str}") from err
 
-    return Rule.parse_obj(result.as_dict())
+_OFFSET_RE_PATTERN = build_offset_regex()
+_START_END_RE_PATTERN = build_start_end_regex()
```
allenporter, line 278: lets move these constants and `build_offset_regex` and `build_start_end_regex` above the `TzParser` declaration.

ical/tzif/tz_rule.py
```diff
@@ -200,64 +181,98 @@ def default_dst_offset(cls, values: dict[str, Any]) -> dict[str, Any]:
...
+
+def build_offset_regex() -> re.Pattern[str]:
+    """Create a regular expression for the given TZ string. Alternative implementation to pyparsing."""
+    # name
+    result = r"(?P<name>(\<[+\-]?\d+\>|[a-zA-Z]+))"
+    # offset
+    result += (
+        r"((?P<hour>[+-]?\d+)(?::(?P<minutes>\d{1,2})(?::(?P<seconds>\d{1,2}))?)?)?"
+    )
+    return re.compile(result)
```
allenporter, line 258: Does a syntax like this fit?
```suggestion
    return re.compile(
        r"(?P<name>(\<[+\-]?\d+\>|[a-zA-Z]+))"   # name
        r"((?P<hour>[+-]?\d+)(?::(?P<minutes>\d{1,2})(?::(?P<seconds>\d{1,2}))?)?)?"  # offset
    )
```

ical/tzif/tz_rule.py
```diff
@@ -200,64 +181,98 @@ def default_dst_offset(cls, values: dict[str, Any]) -> dict[str, Any]:
...
+
+        return RuleOccurrence(
+            name=match.group("name"), offset=_parse_time(match.groupdict()) or _ZERO
         )
+
+    def _parse_tz_rule_date(self) -> Union[RuleDate, RuleDay, None]:
+        """Parse the TZ string into a RuleDate or RuleDay object."""
+        match = _START_END_RE_PATTERN.match(self._str, self._pos)
+        if not match:
+            return None
```
allenporter, line 232:
```suggestion
        if not (match := _START_END_RE_PATTERN.match(self._str, self._pos)):
            return None
```

ical/tzif/tz_rule.py
```diff
@@ -200,64 +181,98 @@ def default_dst_offset(cls, values: dict[str, Any]) -> dict[str, Any]:
...
+            raise ValueError(
+                f"Unable to parse TZ string: {self._str}. Unexpected end of string at position {self._pos}."
+            )
+
+        return Rule(std=std, dst=dst, dst_start=std_start, dst_end=std_end)
+
+    def _parse_tz_rule_occurrence(self) -> RuleOccurrence | None:
+        """Parse the TZ string into a RuleOccurrence object."""
+        match = _OFFSET_RE_PATTERN.match(self._str, self._pos)
+        if not match:
```
allenporter, line 219:
```suggestion
        if not (match := _OFFSET_RE_PATTERN.match(self._str, self._pos)):
```

ical/tzif/tz_rule.py
```diff
@@ -200,64 +181,98 @@ def default_dst_offset(cls, values: dict[str, Any]) -> dict[str, Any]:
...
-        + Opt(Group(onset).set_results_name("dst"))
-        + Opt(
-            ","
-            + Group(tz_date).set_results_name("dst_start")
-            + ","
-            + Group(tz_date).set_results_name("dst_end")
+    def parse(self) -> Rule:
+        """Parse the TZ string into a Rule object."""
+        std = self._parse_tz_rule_occurrence()
+        if not std:
```
allenporter, line 197:
```suggestion
        if not (std := self._parse_tz_rule_occurrence()):
```

ical/tzif/tz_rule.py
```diff
@@ -200,64 +181,98 @@ def default_dst_offset(cls, values: dict[str, Any]) -> dict[str, Any]:
...
+            raise ValueError(f"Unable to parse TZ string: {self._str}")
+
+        dst = self._parse_tz_rule_occurrence()
+        std_start = self._parse_tz_rule_date()
+        std_end = self._parse_tz_rule_date()
+
+        # either both dates are set or both are None
+        if (std_start is None) != (std_end is None):
```
allenporter, line 205: I was staring at the previous lines for awhile trying to understand what would happen if one date failed to parse and the cursor was not advanced, then found this. I don't love that there are side effects in these parse methods as that would not necessarily be expected.  Have altnerative idea for how to deal with the position movement for optional fields that is simpler?

ical/tzif/tz_rule.py
```diff
@@ -24,54 +24,49 @@
...
     sign = 1
     if hour.startswith("+"):
         hour = hour[1:]
     elif hour.startswith("-"):
         sign = -1
         hour = hour[1:]
-    minutes = values.get("minutes", "0")
-    seconds = values.get("seconds", "0")
-    return sign * (int(hour) * 60 * 60 + int(minutes) * 60 + int(seconds))
+    minutes = values.get("minutes") or "0"
```
allenporter, line 65: Why did this change? Can the map now contain `None`?

ical/tzif/tz_rule.py
```diff
@@ -24,54 +24,49 @@
...
     """Convert an offset from [+/-]hh[:mm[:ss]] to a valid timedelta pydantic format.
 
     The parse tree dict expects fields of hour, minutes, seconds (see tz_time rule in parser).
     """
-    if not values:
-        return 0
-    if isinstance(values, (int, str, datetime.timedelta)):
+    if isinstance(values, datetime.timedelta):
         return values
+    if not values:
```
allenporter, line 51: This is not currently tested. Can it happen?

ical/tzif/tz_rule.py
```diff
@@ -24,54 +24,49 @@
...
+        return None
     if not isinstance(values, dict):
-        raise ValueError("time was not parse tree dict, timedelta, string, or int")
+        raise ValueError("time was not parse tree dict or timedelta")
+
     hour = values["hour"]
+    if hour is None:
```
allenporter, line 57:
```suggestion
    if (hour := values["hour"]) is None:
```

allenporter (changes requested):
Thank you appreciate the cleanup. I want to take a closer look at the regexps but thought this was a great start. I want to double check we have sufficient test coverage for these cases etc

allenporter:
I had trouble resolving to trying again in #518 to drop pyparsing pulling in all your changes. Thank you!

PR #514, opened 2025-07-09
Modernize packaging

cdce8p:
- Move static project metadata to `pyproject.toml`
- Adopt PEP 639 license expressions -> removed license classifier
https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files
- Removed explicit `include_package_data = True`. That's the default for `pyproject.toml` configs. Also removed explicit mention of `py.typed`. It's included by default.

Metadata diff
```diff
 ...
-Home-page: https://github.com/allenporter/ical
+Project-URL: Source, https://github.com/allenporter/ical
-Author: Allen Porter
-Author-email: allen.porter@gmail.com
+Author-email: Allen Porter <allen.porter@gmail.com>
-License: Apache-2.0
+License-Expression: Apache-2.0
-Classifier: License :: OSI Approved :: Apache Software License
 ...
```


allenporter:
Yes, definitely didn't apply on it's own. What I can do is merge your PR then run the cruft update which will be more of a no-op. The cruft update has more drift than what you have here.

PR #537, opened 2025-10-23
Fix DUE checking in recurrent todo

### LINKED ISSUE #534: Can't create reccurring Todo with duration
**Describe the bug**
There's no rule against recurrent `Todo` without `due` such Todo should validate.

**To Reproduce**
```python
import ical.todo, ical.types as it
ical.todo.Todo(
                dtstart=dt.date(2025, 1, 1),
                #due=dt.date(2025, 1, 1),
                rrule=ical.types.Recur(
                    freq=ical.types.Frequency.WEEKLY,
                    interval=2,
                    by_weekday=[it.WeekdayValue(it.Weekday.MONDAY)],
                ),
                duration=dt.timedelta(minutes=15),
            )
```
output:
```
ical.exceptions.CalendarParseError: Event must have a due date to be recurring
```

**Expected behavior**
I think in this case it should check for due OR duration.

BTW, should read it "Todo must have..."?

**Additional context**
I want to create a recurrent todo item expressing an item that should take 15 minutes and should be completed sometime during a given day.

Tried making a `Todo` with `dtstart`, `rrule` and `duration` but that fails on missing `due`.
I can't add due, since it's against the RFC to have both DUE and DURATION (even tho it would make sense with DATE values) and it's validated against.
It makes sense to check if a recurrent todo finishes, but I think `duration` should also work.

If the above makes sense, it's a one-liner and I can provide a patch.

rafalkrupinski:
- Fix error message on both DUE and DURATION present
- Delete checking for DUE in `Todo.as_rrule()`
- Remove test cases that check for missing DTSTART with DTSTART present

Since DUE is exclusive with DURATION, by making DUE optional, this...

Fixes #534

### DIFF
--- a/tests/test_todo.py
+++ b/tests/test_todo.py
@@ -92,6 +92,11 @@ def test_dtstart_date_duration_hours_invalid():
         ),
         (
             {
+                "duration": datetime.timedelta(hours=1),
+            }
+        ),
+        (
+            {
                 "start": datetime.datetime(2022, 9, 6, 6, 0, 0),
                 "due": datetime.datetime(
                     2022, 9, 7, 6, 0, 0, tzinfo=zoneinfo.ZoneInfo("America/Regina")
--- a/ical/todo.py
+++ b/ical/todo.py
@@ -330,8 +330,8 @@ class Todo(ComponentModel):
         """
         if not self.rrule and not self.rdate:
             return None
-        if not self.due:
-            raise CalendarParseError("Event must have a due date to be recurring")
+        if not self.due and not self.duration:
+            raise CalendarParseError("Event must have a due date or duration to be recurring")
         return as_rrule(self.rrule, self.rdate, self.exdate, self.dtstart)
 
     _validate_until_dtstart = model_validator(mode="after")(validate_until_dtstart)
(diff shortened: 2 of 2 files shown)


tests/test_todo.py
```diff
@@ -72,32 +72,11 @@ def test_duration() -> None:
...
-                "start": datetime.datetime(2022, 9, 6, 6, 0, 0),
-            }
-        ),
         (
             {
                 "due": datetime.datetime(2022, 9, 6, 6, 0, 0),
             }
         ),
-        (
-            {
```
allenporter: Can we add tests for how these cases do behave w.r.t. recurring TODO? You mentioned duration having an effect here, though not sure that happens as of this PR yet. Any other tests for duration that need to be added or is existing coverage good?

Thank you for your contribution.

allenporter: I see you had a question about this in the issue but i didn't understand the question exactly.

rafalkrupinski:
> Can you add tests for the cases you are deleting from test_validate_rrule_required_fields 

I didn't look at them closely, sorry. Brought them back.

PR #551, opened 2025-11-13
Add tests for Todo.computed_duration and end properties.

rafalkrupinski:
Follow up to previous [PR](https://github.com/allenporter/ical/pull/541#issuecomment-3517310231)

>    end when there is a start date but not due/duration (effectively testing the default)
>    computed_duration when there is no start or duration

Also fixed test function name conflict I've introduced


allenporter:
Thank you!

PR #563, opened 2026-01-02
fix same day compat

sandu-c:


ical/compat/make_compat.py
```diff
@@ -43,6 +44,11 @@ def enable_compat_mode(ics: str) -> Generator[str]:
             timezone_compat.enable_extended_timezones(),
         ):
             yield ics
+    elif prodid and _CALENDAR_LABS_PRODID in prodid:
```
allenporter: I'm OK to enable this for all calendars given as you said it's a common problem. Can you apply in the outer scope and remove the "No compatibility mode needed" debug line?

allenporter:
> Thank you, these tests are very strong. One quick comment and I can get this merged and released for you.

There is a test failure

tests/compat/test_make_compat.py
```diff
@@ -21,11 +21,19 @@ def test_make_compat(filename: pathlib.Path, snapshot: SnapshotAssertion) -> Non
     with filename.open() as ics_file:
         ics = ics_file.read()
         with enable_compat_mode(ics) as compat_ics:
-            assert timezone_compat.is_allow_invalid_timezones_enabled()
-            assert timezone_compat.is_extended_timezones_enabled()
+            # Only check timezone compatibility for Exchange Server files
+            if "office_3" in filename.stem:
```
allenporter: Please change the text fixtures instead of doing this. That is, lets have different tests with different file globs for the test that apply. I don't want an accidental mistake related to the filename to mask this.

allenporter (changes requested):
The tests don't pass and i have some requested changes below.

tests/compat/test_make_compat.py
```diff
@@ -11,50 +12,71 @@
...
-                assert timezone_compat.is_extended_timezones_enabled()
             calendar = IcsCalendarStream.calendar_from_ics(compat_ics)
 
     new_ics = IcsCalendarStream.calendar_to_ics(calendar)
     
     # Normalize DTSTAMP for calendar_labs files since it's auto-generated
-    if "calendar_labs" in filename.stem:
-        import re
-        new_ics = re.sub(r'DTSTAMP:\d{8}T\d{6}Z', 'DTSTAMP:20260103T212907Z', new_ics)
+    new_ics = re.sub(r'DTSTAMP:\d{8}T\d{6}Z', 'DTSTAMP:20260103T212907Z', new_ics)
```
allenporter: Use a fixture like this to set the timestamp used for dtstamp:

```
@pytest.fixture(name="frozen_time", autouse=True)
def mock_frozen_time() -> Generator[FrozenDateTimeFactory, None, None]:
    """Fixture to freeze time to a specific point."""
    with freeze_time("2022-09-03T09:38:05") as freeze:
        with patch("ical.event.dtstamp_factory", new=freeze):
            yield freeze
```

allenporter:
Ah tests are currently failing on CI:
`Re-run pytest with --snapshot-update to delete unused snapshots.`

Should be good to go after that.

allenporter:
I'm still interested in merging this if you'd like to fix the tests. You can mark as ready for review when ready.

PR #565, opened 2026-01-26
Dedupe edits to a single instance of a recurring event

### LINKED ISSUE #278: Revisit RECURRENCE-ID, RANGE, UID

stutrek:
This PR dedupes edited recurring events. Right now if a recurring event is edited, the original shows alongside the edited version, making a duplicate entry. An example of this is even in the existing snapshot data.

This PR does not fully fix #278, but it will fix the issue almost all the time, as many calendar clients delete the old event and create a new one when "this and all recurring events" is chosen instead of "just this event".

This scratches [my personal itch](https://github.com/home-assistant/core/issues/143455).
 
 I have another PR coming that will properly handle recurring events, it will include this PR, but is a much larger change.
 
 In this PR:
 
 - Normalize dates to ensure time zones work
 - group events by their uid
 - dedupe edited events using their start time, as specified in the spec.
 
 There should be no breaking changes.

### DIFF
--- a/ical/calendar.py
+++ b/ical/calendar.py
@@ -19,6 +19,7 @@ from .journal import Journal
 from .types.date_time import TZID
 from .types import RecurrenceId
 from .parsing.property import ParsedProperty
+from .recur_adapter import items_by_uid
 from .timeline import Timeline, calendar_timeline
 from .timezone import Timezone, TimezoneModel, IcsTimezoneInfo
 from .todo import Todo
@@ -153,25 +154,37 @@ class Calendar(ComponentModel):
         items: list[Event] | list[Todo] | list[Journal],
     ) -> None:
         """Reconcile recurrence overrides for a list of components."""
-        # Group by UID
-        by_uid: dict[str, list[Event | Todo | Journal]] = {}
-        for item in items:
-            if item.uid:
-                by_uid.setdefault(item.uid, []).append(item)
+        # Only items with rrule or recurrence_id participate in override
+        # reconciliation. Items without a uid are also excluded.
+        relevant = [
+            item for item in items
+            if item.uid and (item.rrule or item.recurrence_id)
+        ]
+        if not relevant:
+            return
+
+        by_uid = items_by_uid(relevant)
 
         for uid_items in by_uid.values():
-            # Find the parent recurring event (has rrule, no recurrence_id)
+            # Find the parent recurring event (has rrule, no recurrence_id).
+            # Note: a THISANDFUTURE fork has both rrule AND recurrence_id,
+            # so it won't match here — it's a separate recurring series
+            # that doesn't need exdate fixup.
             parent = next(
                 (i for i in uid_items if i.rrule and not i.recurrence_id), None
             )
             if not parent:
                 continue
 
-            # Find edited instances (has recurrence_id, no rrule)
+            # Find edited single instances (has recurrence_id, no rrule).
+            # Items with both rrule and recurrence_id are THISANDFUTURE
+            # forks, not single-instance overrides.
             for item in uid_items:
                 if item.recurrence_id and not item.rrule:
                     exdate = RecurrenceId.to_value(item.recurrence_id)
-                    # Match timezone to parent's dtstart for proper comparison
+                    # Align timezone with parent's dtstart so dateutil
+                    # can match the exdate during rruleset expansion.
+                    # Same pattern as store.py _apply_delete.
                     if (
                         isinstance(exdate, datetime.datetime)
                         and isinstance(parent.dtstart, datetime.datetime)


ical/recur_adapter.py
```diff
@@ -88,21 +88,66 @@ def items_by_uid(items: list[ItemType]) -> dict[str, list[ItemType]]:
...
+                iters.append(
+                    [
+                        SortableItemValue(
+                            item.timespan_of(tzinfo),
+                            item,
+                        )
+                    ]
+                )
+            else:
+                # Recurring item - filter out dates that have been overridden
```
allenporter: Won't this iterate over the entire series inline? that seems like a problem.

stutrek:
This is not reproducible when creating events with the Store API because this library adds EXDATE during edit. Other providers, like Google and Apple, do not use EXDATE, but only RECURRENCE-ID. This is valid in the spec. 

I didn't realize this already worked, but the files I was using were simply missing the field that's used to check. We can just add EXDATE at parse time and it will work as intended. It has a side effect that a serialized ics file will have EXDATE added to recurring events (still valid, just different). I updated the PR to do this.

Here's some proof of the bug, a small calendar created in Google Calendar and shown in Home Assistant. I have included this ics in the repo as test data. It has an event that occurs three times, with the second one edited.

This is what it looks like in Google Calendar:
image

In Home Assistant, it looks like this:
image

Note that the second event is doubled up.

This bug has been reported to HA several times, my heart goes out to everyone that maintains popular open source projects, they... politely... direct people here.

I think this was the whole problem and my other PR is necessary, I will close it.

stutrek:
This screenshot is from the remote calendar integration. The ICS file I included is curl'd from the URL Google gave me to paste into HA.

I went through a few iterations with AI on how to do this, trying to see if there was a way to have this library work with only recurrence-id, but most of them were kind of far reaching and didn't look like a good fit. IMO this PR is the best one, but in case the serialized data is important to you, I made #572. It uses something similar to the initial method, but makes a set for fast lookup.

ical/calendar.py
```diff
@@ -133,4 +134,52 @@ def _propagate_timezones(cls, values: dict[str, Any]) -> dict[str, Any]:
...
+
+    @staticmethod
+    def _reconcile_component_overrides(
+        items: list[Event] | list[Todo] | list[Journal],
+    ) -> None:
+        """Reconcile recurrence overrides for a list of components."""
+        # Group by UID
+        by_uid: dict[str, list[Event | Todo | Journal]] = {}
+        for item in items:
+            if item.uid:
```
allenporter: Is it  (a) correct and (b) faster for this to only look at values with either `rrule` or `recurrence-id`?

ical/calendar.py
```diff
@@ -133,4 +134,52 @@ def _propagate_timezones(cls, values: dict[str, Any]) -> dict[str, Any]:
...
+            self._reconcile_component_overrides(component_list)
+        return self
+
+    @staticmethod
+    def _reconcile_component_overrides(
+        items: list[Event] | list[Todo] | list[Journal],
+    ) -> None:
+        """Reconcile recurrence overrides for a list of components."""
+        # Group by UID
+        by_uid: dict[str, list[Event | Todo | Journal]] = {}
```
allenporter: I think this can be done with `itertools.groupby(items, key=lambda x: x.uid)`

ical/calendar.py
```diff
@@ -133,4 +134,52 @@ def _propagate_timezones(cls, values: dict[str, Any]) -> dict[str, Any]:
...
+        for item in items:
+            if item.uid:
+                by_uid.setdefault(item.uid, []).append(item)
+
+        for uid_items in by_uid.values():
+            # Find the parent recurring event (has rrule, no recurrence_id)
```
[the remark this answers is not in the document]
stutrek: Do you have test data with this situation where we could delete the EXDATE? It's going to be very hard for me to recreate this, the tools I use to make calendars make a new event when you check the "this and future events" checkbox.

stutrek:
Ok, this is pretty hilarious. I asked Opus 4.6 to make a plan addressing these comments, this is what it came back with. If you like this solution more I'll argue with it when I have time, but I think it makes some good points.

image

ical/calendar.py
stutrek: itertools.groupby requires pre-sorted input, so it wouldn't work directly here. It looks like filtering the data and using `items_by_uuid` will work.

stutrek:
I gave the AI the whole spec, it chugged through a few dollars worth of tokens and said we shouldn't do anything for THISANDFUTURE. So I dug in and ended up agreeing. All the cases this library currently works for will continue to work, and the changes today do reduce the scope of the change.

I tried THISANDFUTURE in the big calendar apps, and came away surprised at how bad they are. I understand your passion for calendar libraries.

- Apple - truncates and makes a new event. When you delete the whole event, all single event edits remain on your calendar as bugs and you can't delete them, you'll get notifications about them that you can't stop.
- Google - truncates but cleans up the edited events
- Microsoft - does not support THISANDFUTURE, only editing ALL events; editing an event in the middle causes previous events to update.
- Thunderbird - same as MS

If we were to make test data for this it would be complex to create, and entirely hypothetical. We can't generate failing data from an existing product like we can for the single event bug.

I have no strong opinion on which PR is better, but my entire family begs you accept one of these, or add the test data to #513 and accept it if the test passes.

PR #569, opened 2026-01-31
Fix graceful degradation for unknown VALUE parameters (#567)

### LINKED ISSUE #567: Crashes with "special characters"
**Describe the bug**
I am switching from Gmail to Proton and have imported all calendar events. In HA I have used the "Remote Calendar" integration that uses ical 12.1.2.
When running I get the following error:
Error parsing calendar information: Failed to parse calendar EVENT component: Value error, Failed to validate: IBE150 NAME NAMESON\n A283\n as str, due to: (['Property parameter specified unsupported type: IBE150 NAME NAMESON^n A283^n'])
Error parsing calendar information: Failed to parse calendar EVENT component: Value error, Failed to validate: IBE150 NAME NAMESON\n A285\n as str, due to: (['Property parameter specified unsupported type: IBE150 NAME NAMESON^n A285^n'])

**To Reproduce**
Extracted from the ics file:
BEGIN:VEVENT
SUMMARY;VALUE="IBE150 NAME NAMESON^n A285^n":IBE150 NAME NAMESON\n A2
 85\n
UID:20100913T082504-305640@158.38.83.24
DTSTART;TZID=Europe/Paris:20100913T081500
DTEND;TZID=Europe/Paris:20100913T100000
SEQUENCE:0
STATUS:CONFIRMED
DTSTAMP:20260130T184242Z
END:VEVENT

**Expected behavior**
Should work

**Additional context**
I checked the file with a ical validator and it cleared.

fiLLLip:
Implements RFC 5545 compliant handling of unknown VALUE parameter types by falling back to TEXT parsing instead of raising ValueError.

Problem:
- Calendar files from Proton Mail and other sources use custom VALUE parameters (e.g., VALUE="IBE150 NAME NAMESON^n A285^n")
- The library was strictly rejecting any VALUE type not in the hardcoded registry of 17 standard types
- This caused ValueError crashes when parsing otherwise valid calendars

Solution:
- Modified component.py to gracefully fall back to TEXT parsing for unknown VALUE types
- Added warning logs when fallback occurs for debugging
- Maintains data integrity while improving interoperability

Changes:
- ical/component.py: Replace ValueError with TEXT fallback in _parse_single_property() method
- tests/test_component.py: Add 4 comprehensive tests for unknown VALUE types including real Proton Mail examples
- tests/types/test_date_time.py: Update tests to reflect new graceful degradation behavior

RFC 5545 Compliance:
Section 3.2.20 states: "Applications MUST preserve the value data for x-name and iana-token values that they don't recognize without attempting to interpret or parse the value data."

This change ensures compliance by preserving unknown VALUE types through TEXT fallback rather than rejection.

Fixes #567


ical/component.py
```diff
@@ -256,9 +256,19 @@ def _parse_single_property(cls, field_type: type, prop: ParsedProperty) -> Any:
...
+
+            # Graceful degradation: fall back to TEXT parsing for unknown VALUE types
+            _LOGGER.warning(
+                "Property '%s' has unsupported VALUE type '%s', falling back to TEXT parsing",
+                prop.name,
+                value_type,
+            )
+            if text_parser := DATA_TYPE.parse_parameter_by_name.get("TEXT"):
+                return text_parser(prop)
+            # If TEXT parser is not available (should never happen), fall through
```
allenporter: if it should never happen then maybe still raising ValueError is reasonable or just returning bare value directly?

allenporter:
Thank you for the improvement, i agree with your reading of the spec. 

One question i have is about serialization: Does this work to round trip the output .ics? I think we may not be preserving the value on serialize, but not sure

One thing I like to do is also add the .ics files for realistic files in the `tests/testdata/` which I think will exercise some of the round tripping if I remember right. I do like that you have the inline real world example in the test, but its also nice to have a real-ish .ics file too.

allenporter:
> I'll answer this first 😄 You are correct, we do not preserve the fields in a round trip. As for test data, I have tested on my own real life calendar and won't be pushing that as it's "too sensitive" (may not be the case but it's personal at least 👯 )

Yes, agree, I don't want you to put personal information in this repo.

allenporter:
Thanks! 👍🏼

PR #572, opened 2026-02-02
Match edits for recurring events by RECURRENCE-ID

stutrek:
This is an alternative to #565. See the full message and justification there.

Instead of mutating data as it comes in, this changes the merge and expansion method to:

- group events by UID 
- creates a set of override dates
- skips overridden dates

This has a slight efficiency loss over #565 especially if the model is iterated many times, and does not use dateutil for comparison. It to me seems less integrated into the library, the upside is that serialized ics files will be the same as the input.

Issue #278

### DIFF
--- a/tests/examples/test_ics_examples.py
+++ b/tests/examples/test_ics_examples.py
@@ -44,6 +44,7 @@ def test_recurring_with_single_change() -> None:
 
     # Feb 1 - original title
     assert events[0].summary == "Initial Title"
+    assert events[0].recurrence_id == "20260201T100000"
     assert events[0].dtstart == datetime.datetime(2026, 2, 1, 10, 0, tzinfo=tz)
 
     # Feb 2 - edited instance with RECURRENCE-ID
@@ -53,4 +54,5 @@ def test_recurring_with_single_change() -> None:
 
     # Feb 3 - original title
     assert events[2].summary == "Initial Title"
+    assert events[2].recurrence_id == "20260203T100000"
     assert events[2].dtstart == datetime.datetime(2026, 2, 3, 10, 0, tzinfo=tz)


ical/recur_adapter.py
```diff
@@ -77,32 +123,78 @@ def build() -> ItemType:
 
 
 def items_by_uid(items: list[ItemType]) -> dict[str, list[ItemType]]:
-    items_by_uid: dict[str, list[ItemType]] = {}
+    result: dict[str, list[ItemType]] = {}
```
allenporter: Does this need to change? I'd say revert anything that is not a meaningful change from this PR to make review easier.

ical/recur_adapter.py
```diff
@@ -77,32 +123,78 @@ def build() -> ItemType:
...
+        parent = next(
+            (i for i in uid_items if i.rrule and not i.recurrence_id), None
+        )
+        parent_tzinfo = (
+            parent.dtstart.tzinfo
+            if parent and isinstance(parent.dtstart, datetime.datetime)
+            else None
+        )
+
+        # Collect override dates from edited instances as a frozenset for O(1)
```
allenporter: Would it work to make this work in terms of the found recurrence ids?
- Collect all the `item.recurrence_id` in one iteration
- As we're iterating over events and we see `parent` skip anything matching the recurrence id.

I think matching by recurrence_id (which we generate anyway as we iterate in the recur adapter) might be easier to match on than trying to normalize the dates -- assuming that works.

stutrek:
> I kind of like this one... Assuming it's not super slow.?
> 
> We do have some tests of iter speed fwiw if you want to do a before and after with a case that exercises this:
> 
> https://github.com/allenporter/ical/blob/017699acd7e9a9a0dc316087450cab4862e9352a/tests/test_timeline.py#L47
> 
> 
> https://github.com/allenporter/ical/blob/017699acd7e9a9a0dc316087450cab4862e9352a/tests/test_iter.py#L87

Sure, I'll use this one. But I do want to warn you, on the flip side of open source libraries, I only have another round or two of review in me.

allenporter:
I went ahead and made the commits I suggested. Thanks @stutrek 👍🏼

PR #573, opened 2026-02-07
Accept `DATE` values with midnight time suffix

arrizer:
Accept `DATE` values that include a midnight time suffix (`YYYYMMDDT000000` or `YYYYMMDDT000000Z`).

Some calendar generators emit invalid `VALUE=DATE` values with a midnight time suffix. This is common in real-world ICS feeds. Strict parsing rejects these files even though the intent is unambiguous. Treating midnight time suffixes as dates improves interoperability without impacting valid `DATE-TIME` inputs.

### DIFF
--- a/ical/types/date.py
+++ b/ical/types/date.py
@@ -13,6 +13,8 @@ from .data_types import DATA_TYPE
 _LOGGER = logging.getLogger(__name__)
 
 DATE_REGEX = re.compile(r"^([0-9]{8})$")
+# Some generators emit invalid DATE values with a midnight time suffix.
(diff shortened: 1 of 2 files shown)


allenporter:
Thanks again for the contribution and raising this! I've resolved this issue by implementing a date compatibility layer under the ical.compat module (matching the feedback here). MALFORMED `DATE` values with time suffixes will now be silently stripped and parsed successfully when compatibility mode is enabled.

You can track the implementation details in #625. Closing this PR in favor of that solution. Thank you.

Issue #546, opened 2025-11-11
Feature Improvement Extend CLASS parameter


RalfKraemer1:
According to https://datatracker.ietf.org/doc/html/rfc5545#section-3.8.1.3 
the CLASS parameter  is not locked to "PUBLIC" / "PRIVATE" / "CONFIDENTIAL"

That attribute makes it seem like this can support any other arbitrary values. This will need a feature improvement in the ical library to support this.

Please improve the Library.

allenporter:
Agree this is worth prioritizing.

PR #575, opened 2026-02-10
Added support for non-standard CLASS values (#546).

teedog:
Added support for non-standard CLASS values (#546).

### DIFF
--- a/ical/types/const.py
+++ b/ical/types/const.py
@@ -31,6 +31,4 @@ class Classification(str, enum.Enum):
     @classmethod
     def __parse_property_value__(cls, prop: ParsedProperty) -> Self | None:
         """Parse value into enum (accepts any string)."""
-        if prop.value is None:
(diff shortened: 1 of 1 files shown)


ical/types/const.py
```diff
@@ -16,10 +16,21 @@ class Classification(str, enum.Enum):
...
     @classmethod
     def __parse_property_value__(cls, prop: ParsedProperty) -> Self | None:
-        """Parse value into enum."""
-        try:
-            return cls(prop.value)
-        except ValueError:
+        """Parse value into enum (accepts any string)."""
+        if prop.value is None:
             return None
+        return cls(prop.value)
```
allenporter: Lint error on CI says this is not reachable:

```suggestion
        return cls(prop.value)
```

allenporter:
Thanks @teedog 👍🏼

Issue #536, opened 2025-10-19
`test_make_compat.py`'s `test_make_compat` method fails when round-tripping an ICS with `TZID=Pacific Standard Time`


ChrisCarini:
**Describe the bug**
A clear and concise description of what the bug is, and the version of the ics library used.

It appears that an ICS file with something like `DTSTART;TZID=Pacific Standard Time:20241018T000000` (noting the `Pacific Standard Time` tz) will cause this lib to fail the 'round-trip' of converting the ICS file into a `Calendar` object, and then back to an ICS (`str`) via `IcsCalendarStream.calendar_to_ics(calendar)` will change the line to `DTSTART;TZID=PST:20241018T000000`

**To Reproduce**

See https://github.com/allenporter/ical/pull/535

<img width="1860" height="1200" alt="Image" src="https://github.com/user-attachments/assets/0f4c35a6-73e1-4904-a4b0-7bd8797fd2aa" />

**Expected behavior**

I would expect at the very least that the ICS str coming from `IcsCalendarStream.calendar_to_ics(calendar)` would be re-parsable by this lib.

I understand that `Pacific Standard Time` is not a 'standard' TZ, and is part of the recent compatibility that was added for Office 365, but "most-ideally", it'd be great if `IcsCalendarStream.calendar_to_ics(calendar)` could return a value that is near-identical to the original ICS file that was passed in.

**Additional context**

I found this when trying to do some testing/debugging for https://github.com/home-assistant/core/issues/148315 - likely tangential and not directly related, just but 'linking' for self-posterity.

ChrisCarini:
Hey @allenporter - sorry for the delay!

I see `Pacific Standard Time`:
```
BEGIN:VTIMEZONE
TZID:Pacific Standard Time
BEGIN:STANDARD
DTSTART:16010101T020000
TZOFFSETFROM:-0700
TZOFFSETTO:-0800
RRULE:FREQ=YEARLY;INTERVAL=1;BYDAY=1SU;BYMONTH=11
END:STANDARD
BEGIN:DAYLIGHT
DTSTART:16010101T020000
TZOFFSETFROM:-0800
TZOFFSETTO:-0700
RRULE:FREQ=YEARLY;INTERVAL=1;BYDAY=2SU;BYMONTH=3
END:DAYLIGHT
END:VTIMEZONE
```

Full excerpt (lines 6-134 of my remote ics; the entirety of `VTIMEZONE`) below:
```
BEGIN:VTIMEZONE
[...]

PR #591, opened 2026-04-26
Cache TzInfo instances and per-year DST boundaries & add regression test for Office 365 extended-timezone perf

ChrisCarini:
## Claude's Assessment

_**Note:** Below is using Opus 4.7(1M) model._

### Root cause

The slowness is in the `ical` library, specifically `ical/tzif/timezoneinfo.py`. Two compounding issues hit Office 365 calendars
hard:

1. `read_tzinfo()` was not cached. Every parsed datetime with `TZID=Pacific Standard Time` got its own fresh `TzInfo` instance. CPython's datetime comparison fast‑paths `mytz is ottz`; with every event having a unique tzinfo, every comparison falls back to calling `utcoffset()` on both sides.
2. `TzInfo.dst()` rebuilt two `dateutil.rrule` objects on every call to compute that year's DST start/end (`timezoneinfo.py:186-188`).

For your `calendar.ics` (1.82 MB, 2543 events, all referencing `Pacific Standard Time`), `materialize_timeline` calls `intersects` / heap `__lt__` ~3000+ times during the merge. Each comparison fans out into ~58k `utcoffset` calls and ~115k dateutil rrule constructions. On M1 that's 22s; on a Pi 4 it scales to 2-5 minutes — exactly the symptom in
https://github.com/home-assistant/core/issues/148315.

Issue/PR allenporter/ical#536/535 is a _different_, related issue (round-trip serialization of unknown VTIMEZONE properties). Allen has since added proper `extras` support for that.

### Fix (verified)

Two edits in `ical/ical/tzif/timezoneinfo.py`:
- Cache `_read_tzinfo_resolved(resolved_key)` so all events with the same `TZID` share one `TzInfo` instance. The cache is keyed on the resolved IANA name, so when compat mode is off, non-IANA keys like `"Pacific Standard Time"` still correctly raise `TimezoneInfoError`.
- Memoize per-year DST boundaries on the `TzInfo` instance.

Results on your real `calendar.ics`

|  | materialize_timeline |
|---|-------------------|
| Baseline          | 22.13s             |
| `read_tzinfo` cached | 5.37s           |
| Both fixes        | 0.21s (≈100×)      |

Full ical test suite: 1278 passed, 26 skipped, 0 failed.

## Tests added by Claude

The new test in `tests/test_timeline.py::test_materialize_timeline_office365_extended_timezone_perf`:
- Builds a synthetic Microsoft Exchange ICS with a `Pacific Standard Time` VTIMEZONE and 1500 events all using `TZID=Pacific Standard Time`.
- Parses with `enable_compat_mode` and runs `materialize_timeline` over a 365-day window.
- Asserts elapsed < 1.0s.

Verified behaviour: passes with the fix (~25ms on M1) and fails without it (~8s — well above the 1s budget).

### DIFF
--- a/tests/test_timeline.py
+++ b/tests/test_timeline.py
@@ -2,6 +2,7 @@
 
 import datetime
 import random
+import time
 import zoneinfo
 from typing import Any
 from unittest.mock import patch
@@ -9,6 +10,8 @@ from unittest.mock import patch
 import pytest
 
 from ical.calendar import Calendar
+from ical.calendar_stream import IcsCalendarStream
+from ical.compat import enable_compat_mode
 from ical.event import Event
 from ical.journal import Journal
 from ical.types.recur import Recur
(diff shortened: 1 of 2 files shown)


tests/test_timeline.py
```diff
@@ -211,6 +214,90 @@ def test_benchmark_materialize_timeline() -> None:
...
+                f"DTEND;TZID=Pacific Standard Time:{end.strftime('%Y%m%dT%H%M%S')}",
+                f"SUMMARY:Event {i}",
+                "END:VEVENT",
+            ]
+        )
+    lines.append("END:VCALENDAR")
+    return "\r\n".join(lines) + "\r\n"
+
+
+def test_materialize_timeline_office365_extended_timezone_perf() -> None:
```
allenporter, line 259: Similar to the comment in the other test please use pytest-benchmark and post the results. You can parameterize the benchmark using different sized calendars like we do in the other tests and it can give you even more detailed results about how the time changes with the size of the calendar.

tests/test_timeline.py
```diff
@@ -211,6 +214,90 @@ def test_benchmark_materialize_timeline() -> None:
...
+        stop=now + datetime.timedelta(days=365),
+        max_number_of_events=20,
+    )
+    # Drain the iterator so the heap merge actually runs.
+    list(materialized)
+    elapsed = time.perf_counter() - start
+
+    # On the unfixed code path this takes 10+ seconds even on fast hardware;
+    # the cap is intentionally generous so it does not flake on slow CI.
+    assert elapsed < 1.0, (
```
allenporter, line 295: this seems inappropriate to have in a test running on github CI

tests/test_timeline.py
```diff
@@ -211,6 +214,90 @@ def test_benchmark_materialize_timeline() -> None:
     )
 
 
+def _office365_pst_calendar(num_events: int) -> str:
+    """Build a synthetic Office 365 ICS string with N PST-tagged events."""
+    lines = [
```
allenporter, line 219: Use a multi line string rather than a list of strings like this:

https://github.com/allenporter/ical/blob/7c1be0ae8bd135feab0ba9142d7bc291ed0f303f/tests/test_calendar.py#L290

allenporter (changes requested):
Can you ask you PR to update the PR description to not be a record of your conversation with AI and only focus on the actual change being made here? e.g. commenting on extras stuff is not relevant to this PR.

ChrisCarini:
> Can you ask you PR to update the PR description to not be a record of your conversation with AI and only focus on the actual change being made here? e.g. commenting on extras stuff is not relevant to this PR.

Yeah I can remove them. Fwiw, I manually typed both PR descriptions, intentionally including the summary of each AI for posterity (and others' interest in the respective analysis) sake.

allenporter:
Let me know when you're ready for another look by marking ready for review. Thanks for the contribution.

ChrisCarini:
Closing in favor of https://github.com/allenporter/ical/pull/592

Issue #593, opened 2026-04-26
HA core #148315 ical focused discussion


ChrisCarini:
**Describe the bug**
A discussion was started around remote calender fetching slowness in https://github.com/home-assistant/core/issues/148315 - this issue is to have a more focused discussion on the `ical` aspects of where we're current suspecting the issue may be.

**To Reproduce**

In HA, load a large remote calendar with many events - HA will hang.

**Expected behavior**

A large remote calendar is able to be loaded in a reasonable amount of time

ChrisCarini:
Hi @allenporter - I wanted to start a more focused issue on the `ical` project for the issues that have been previously discussed in the linked HA core issue.

Due to sheer lack of time, I haven't had much time to pick up all the notes and state I was in and try to debug further. **BUT**, I did have a brilliant idea to just throw AI at the two repos (HA core and `ical`), the issue and some additional context I had saved in my notes. I cloned 2 separate copies of each, and fed the same prompts to both Claude and Copilot :copilot: (both using Opus 4.7 1M model), and got the below two draft PRs:

- https://github.com/allenporter/ical/pull/591
- https://github.com/allenporter/ical/pull/592

They both identified the same issue (without context from one another), and I've looked over the code and it seems plausible. 

What do you think? If you feel similar, I can clean these up to your liking, and we can try getting one of these integrated into core. Let me know!

allenporter:
I would say:
(1) look at `pytest-benchmark` and reproduce the problem. This can be it's own PR to get that checked in showing poor results at various calendar sizes.
(2) make a change and see how it impacts the benchmark and send a PR with the results attached.

Thanks. I left a version of these comments on the other PRs since I didn't see this since they all came in at once.

allenporter:
I think we can consider this closed given the recent excellent performance optimization contributions. We can re-open or open a new issue to track further issues.

allenporter:
Resolved by PR #592 and PR #611, which introduced caching for DST transitions in TzInfo.dst() and observance transitions in Timezone.get_observance().

PR #611, opened 2026-06-27
Cache observance transitions in Timezone.get_observance

FuNK3Y:
## Description

`Timezone.get_observance()` re-expands the observance recurrence **from the start on every call**. For Microsoft Exchange / Office 365 "Customized Time Zone" `VTIMEZONE`s, the observance `DTSTART` is in year **1601** with an unbounded `YEARLY` rule, so each offset lookup walks ~400+ years of yearly transitions. A calendar with many events triggers this per event.

Importing a simple Outlook.com calendar in home assistant using the remote calendar blew away my whole Home Assistant. I asked Claude for the fix, and this one looks pretty persuasive and reduced the processing time significantly (17s => 0.005s)

## Root cause

`get_observance` calls `self._observances()` — which builds fresh `RecurIterable(observance.as_ruleset())` iterators and expands them from `DTSTART` (1601) — and walks it linearly to the queried value, **with no caching**, on every call.

This is distinct from the path fixed in #592: that cached `TzInfo.dst()` for timezones that **map to a named `TzInfo`** (e.g. `Pacific Standard Time`). `"Customized Time Zone"` has no canonical name, so it stays an embedded timezone resolved through `Timezone.get_observance` and was not covered.

## Fix

Cache the observance transitions (sorted ascending), extending the cache lazily only as far as the highest datetime queried, then `bisect` it. The recurrence is expanded once instead of per call; repeated lookups become `O(log n)`.

## Results

Measured against the real 81-event Outlook calendar (`ical` parse + `materialize_timeline`):

| | before | after |
| --- | --- | --- |
| `materialize_timeline` | 17.6s | 0.005s |

Outputs are byte-for-byte identical: 9 materialized events and an 864-point UTC-offset sweep (1995–2030, including DST boundaries) match the unpatched library exactly.

## Tests

- `test_get_observance_cache_avoids_repeated_expansion` — deterministic guard (mirrors the `as_rrule` counting test from #592): hammers `get_observance` 4,000 times across two years and asserts `Timezone._observances` is expanded `<= 4` times, plus checks summer/winter offsets are correct.
- `test_benchmark_get_observance_repeated_calls` — `pytest-benchmark` perf regression test for the embedded-VTIMEZONE path.

Full suite passes (`1283 passed`); `ruff`, `ruff format`, `ty`, and `codespell` are clean.

Refs #593. Complements #592.

While I am clearly out of the depth when it comes to the logic. I am happy to invest time & token to refactor and test to your liking.

### DIFF
--- a/ical/iter.py
+++ b/ical/iter.py
@@ -17,6 +17,7 @@ backwards incompatibility due to the internal nature.
 
 from __future__ import annotations
 
+import bisect
 import datetime
 import heapq
 import logging
@@ -42,6 +43,7 @@ __all__ = [
     "RecurIterable",
     "ItemAdapter",
     "LazySortableItem",
+    "CachedTransitionTimeline",
 ]
 
 _LOGGER = logging.getLogger(__name__)
@@ -437,3 +439,54 @@ def as_rrule(
         rdate,
         exdate,
     )
+
+
+class CachedTransitionTimeline(Generic[T]):
+    """A lazily materialized timeline over sorted ``(onset, item)`` transitions.
+
+    Given a callable returning a sorted iterable of ``(onset, item)`` pairs,
+    ``active_at`` returns the item of the latest transition whose onset is at or
+    before the queried value. Transitions are materialized lazily and appended
+    to a cache only as far as the highest value queried so far, so an unbounded
+    recurrence (such as a timezone observance rule starting in year 1601) is
+    expanded once instead of on every lookup.
+    """
+
+    def __init__(
+        self,
+        transitions: Callable[[], Iterable[tuple[Any, T]]],
+    ) -> None:
+        """Initialize CachedTransitionTimeline with a transition source factory."""
+        self._transitions = transitions
+        self._iter: Iterator[tuple[Any, T]] | None = None
+        self._cache: list[tuple[Any, T]] = []
+        self._exhausted = False
+
+    def active_at(self, value: Any) -> T | None:
+        """Return the item active at ``value`` (latest onset <= value)."""
+        # Extend the cache only when the request reaches past what has already
+        # been materialized; the iterator is append-only and never restarted.
+        if not self._exhausted and (not self._cache or self._cache[-1][0] <= value):
+            if self._iter is None:
+                self._iter = iter(self._transitions())
+            while not self._exhausted and (
+                not self._cache or self._cache[-1][0] <= value
+            ):
+                try:
+                    self._cache.append(next(self._iter))
+                except StopIteration:
+                    self._exhausted = True
+        index = bisect.bisect_right(self._cache, value, key=lambda item: item[0])
+        if index == 0:
+            return None
+        return self._cache[index - 1][1]
+
+    def __deepcopy__(self, memo: dict[int, Any]) -> CachedTransitionTimeline[T]:
+        """Return an independent copy with an empty cache.
+
+        ``dateutil.rrule`` iterators hold non-copyable state, so the live
+        iterator and cache are not copied. The transition factory is shared (it
+        yields the same immutable transitions) and the copy rebuilds its cache
+        lazily on demand.
+        """
+        return CachedTransitionTimeline(self._transitions)
--- a/ical/iter.py
+++ b/ical/iter.py
@@ -444,31 +444,32 @@ def as_rrule(
 class CachedTransitionTimeline(Generic[T]):
     """A lazily materialized timeline over sorted ``(onset, item)`` transitions.
 
-    Given a callable returning a sorted iterable of ``(onset, item)`` pairs,
-    ``active_at`` returns the item of the latest transition whose onset is at or
-    before the queried value. Transitions are materialized lazily and appended
-    to a cache only as far as the highest value queried so far, so an unbounded
-    recurrence (such as a timezone observance rule starting in year 1601) is
-    expanded once instead of on every lookup.
+    Given a re-iterable source of sorted ``(onset, item)`` pairs, ``active_at``
+    returns the item of the latest transition whose onset is at or before the
+    queried value. Transitions are materialized lazily and appended to a cache
+    only as far as the highest value queried so far, so an unbounded recurrence
+    (such as a timezone observance rule starting in year 1601) is expanded once
+    instead of on every lookup. The source must be re-iterable, since a fresh
+    iterator is taken from it after a deep copy.
     """
 
     def __init__(
         self,
-        transitions: Callable[[], Iterable[tuple[Any, T]]],
+        transitions: Iterable[tuple[datetime.datetime | datetime.date, T]],
     ) -> None:
-        """Initialize CachedTransitionTimeline with a transition source factory."""
+        """Initialize CachedTransitionTimeline with a sorted transition source."""
         self._transitions = transitions
-        self._iter: Iterator[tuple[Any, T]] | None = None
-        self._cache: list[tuple[Any, T]] = []
+        self._iter: Iterator[tuple[datetime.datetime | datetime.date, T]] | None = None
+        self._cache: list[tuple[datetime.datetime | datetime.date, T]] = []
         self._exhausted = False
 
-    def active_at(self, value: Any) -> T | None:
+    def active_at(self, value: datetime.datetime | datetime.date) -> T | None:
         """Return the item active at ``value`` (latest onset <= value)."""
         # Extend the cache only when the request reaches past what has already
         # been materialized; the iterator is append-only and never restarted.
         if not self._exhausted and (not self._cache or self._cache[-1][0] <= value):
             if self._iter is None:
-                self._iter = iter(self._transitions())
+                self._iter = iter(self._transitions)
             while not self._exhausted and (
                 not self._cache or self._cache[-1][0] <= value
             ):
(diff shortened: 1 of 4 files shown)


tests/test_timezone.py
```diff
@@ -216,3 +218,99 @@ def test_clear_old_dtstamp(mock_prodid: Generator[None, None, None]) -> None:
...
+    winter = timezone.get_observance(datetime.datetime(2024, 1, 15, 12, 0, 0))
+    assert summer is not None
+    assert summer.observance.tz_offset_to.offset == datetime.timedelta(hours=2)
+    assert winter is not None
+    assert winter.observance.tz_offset_to.offset == datetime.timedelta(hours=1)
+
+    original = Timezone._observances
+    call_count = 0
+
+    def counting_observances(self: Timezone) -> object:
```
allenporter: Let's not have a test muck with internal details of the class. Given this is a performance fix, lets just rely on the benchmark to measure performance -- and leave tests to measure correctness of the code.

ical/timezone.py
```diff
@@ -249,12 +263,23 @@ def get_observance(self, value: datetime.datetime) -> _ObservanceInfo | None:
         """Return the specified observance for the specified date."""
         if value.tzinfo is not None:
             raise ValueError("Start time must be in local time format")
-        last_observance_info: _ObservanceInfo | None = None
-        for dt_start, observance_info in self._observances():
-            if dt_start > value:
-                return last_observance_info
-            last_observance_info = observance_info
-        return last_observance_info
+        cache = self._observance_cache
```
allenporter: My understanding is that this is a class level variable, so it's shared across instances of this class, is that right? So given that, this solution is not safe/correct, right? It's setting a global observance for any date time independent of the actual timezone.

Generally, I am familiar with the performance problem here given the office calendar custom timezones start so early, so supportive on collaborating on a fix.

allenporter (changes requested):
Appreciate the diagnosis and proposed solution, though i have some comments about it below. i can get more prescriptive on solutions, but wanted to know how much you and claude want to lean in...

tests/test_timezone.py
FuNK3Y: Agreed — dropped that test in 8064c4a. It now relies on the benchmark for the performance dimension and a plain correctness test (`test_customized_timezone_observance_offsets`) that just checks resolved offsets, no patching of internals.

ical/timezone.py
FuNK3Y: Good thing to double-check. It reads like a class attribute, but `PrivateAttr(default_factory=list)` is pydantic's idiom for **per-instance** mutable state: the factory runs per object and the value lives in `self.__pydantic_private__`, so each `Timezone` gets its own cache — nothing is shared across instances or timezones.

Verified directly with two different custom zones (EU `+1/+2` and US-Eastern `-5/-4`):

```
eu cache id: …551808   us cache id: …058048   same object? False
eu summer: 2:00:00   us summer: -4h     eu winter: 1:00:00   us winter: -5h
```

I added a comment in 8064c4a calling out the per-instance semantics so it's not a trap for the next reader. Happy to switch to a different caching approach if you'd prefer one — and yep, glad to lean in on this.

FuNK3Y:
So, I did another pass with Claude with your early comments. Be as prescriptive as you feel necessary. While I'd love to have this fixed the maintainability of the overall library is what matter most !

ical/timezone.py
allenporter: OK thanks for clarifying, sounds good.

I'd like to adjust the abstraction and caching approach before merging:

1. **Separate concerns in `iter.py`:** Let's extract the caching logic out of `timezone.py` into a generic `CachedTransitionTimeline` class at the end of `ical/iter.py` (similar to the other timeline helpers there). It wraps any sorted transition iterable of `(onset, item)` pairs and exposes an `active_at(value)` method. It lazily materializes and caches transitions only up to the queried datetime.
2. **Simplified timezone state:** With that helper, `Timezone` only needs a single private attribute to manage its cache:
   ```python
   _observance_timeline: Optional[CachedTransitionTimeline["_ObservanceInfo"]] = (
       PrivateAttr(default=None)
   )
   ```
3. **Extend the cache, don't reset it (and keep it deepcopy-safe):** Now that the caching is isolated in `CachedTransitionTimeline`, we can also do the optimization to optimize the cache extension:
   * Instead of clearing the cache list on a miss, we can lazily append new transitions.
   * You may need to also support `__deepcopy__` though (since `dateutil.rrule` generators contain non-copyable locks))


WDYT?

allenporter (changes requested):

ical/timezone.py
FuNK3Y: Done in 64b8354 — implemented all three:

1. **`CachedTransitionTimeline` in `iter.py`** — generic over `(onset, item)`, exposes `active_at(value)`, materializes transitions lazily and caches up to the highest value queried.
2. **`Timezone`** now holds a single private attr:
   ```python
   _observance_timeline: Optional[CachedTransitionTimeline["_ObservanceInfo"]] = (
       PrivateAttr(default=None)
   )
   ```
   `get_observance` just lazily constructs it from `self._observances` and delegates to `active_at`.
3. **Extend, don't reset** — the cache is append-only and the iterator is never restarted. For deepcopy-safety I added `CachedTransitionTimeline.__deepcopy__`: since `dateutil` iterators carry non-copyable state, the copy doesn't clone the live iterator/cache — it shares the (pure) transition factory and rebuilds lazily. Verified `copy.deepcopy` of a warmed `Timezone` and the real `IcsTimezoneInfo.__deepcopy__` path both still resolve correct offsets.

Added unit tests for the helper (lookup ordering, single lazy consumption, deepcopy independence) in `tests/test_iter.py`. Full suite green (1286), ruff/ty/codespell clean. `materialize_timeline` on the original Office 365 calendar stays at ~0.005s.

Thanks for the clear direction — happy to tweak the abstraction further if you'd like.

FuNK3Y:
Here it is - taking your comments into account.

ical/iter.py
```diff
@@ -437,3 +439,54 @@ def as_rrule(
...
+    ``active_at`` returns the item of the latest transition whose onset is at or
+    before the queried value. Transitions are materialized lazily and appended
+    to a cache only as far as the highest value queried so far, so an unbounded
+    recurrence (such as a timezone observance rule starting in year 1601) is
+    expanded once instead of on every lookup.
+    """
+
+    def __init__(
+        self,
+        transitions: Callable[[], Iterable[tuple[Any, T]]],
```
allenporter: I think this can accept an Iterable here rather than a callable? (and update the caller). I consider an Iterable already to have the similar properties (e.g. its kind of like a Callable to get back an iterator i think)

ical/iter.py
```diff
@@ -437,3 +439,54 @@ def as_rrule(
...
+    ) -> None:
+        """Initialize CachedTransitionTimeline with a transition source factory."""
+        self._transitions = transitions
+        self._iter: Iterator[tuple[Any, T]] | None = None
+        self._cache: list[tuple[Any, T]] = []
+        self._exhausted = False
+
+    def active_at(self, value: Any) -> T | None:
```
allenporter: Can we do better than `Any` here and in the constructor? If needed could limit to date/time types (i assume making these sortable items isn't worth it)

ical/iter.py
```diff
@@ -437,3 +439,54 @@ def as_rrule(
         rdate,
         exdate,
     )
+
+
+class CachedTransitionTimeline(Generic[T]):
```
allenporter, line 444: I know before i said not to test the details of the internals but perhaps now that its a separate class (and i asked you to make it more complex) a couple extra corner case tests could be useful (still without digging into internals, but just verifying correctness of the cache in different request order scenarios)

allenporter (changes requested):
This is great, much appreciated. One last round and i think we're good to merge.

ical/iter.py
FuNK3Y: Done in f77d73d — now takes `Iterable[tuple[onset, item]]` and the caller passes `self._observances()` directly.

One subtlety worth recording: the source must be **re-iterable**, because `__deepcopy__` can't copy it. Every `dateutil` rrule carries a `threading.Lock`, so deep-copying the transition source fails with `cannot pickle '_thread.lock' object`. So the copy *shares* the source and takes its own fresh `iter()` — which is safe precisely because `MergedIterable` is re-iterable (I verified independent/concurrent traversals stay consistent). I noted that contract in the docstring.

ical/iter.py
FuNK3Y: Done — typed the onset as `datetime.datetime | datetime.date` (matching `_observances()`'s element type) in the constructor, `active_at`, the cache and the iterator. Left the item as the existing `Generic[T]`. Agreed that making these full `SortableItem`s isn't worth it here.

ical/iter.py
FuNK3Y: Added in f77d73d — a parametrized `test_cached_transition_timeline_request_order` covering out-of-order, descending, ascending, repeated and interleaved-extremes request sequences, plus an onset-equality assertion in `test_cached_transition_timeline_active_at`. All via the public `active_at` (no internals). I also kept a lazy-consumption test that asserts the source is iterated once and only pulled far enough to answer the query, and a deepcopy-independence test — both through public behavior. Full suite 1291 green.

FuNK3Y:
Changes implemented.

ical/iter.py
allenporter: > One subtlety worth recording: the source must be **re-iterable**, because `__deepcopy__` can't copy it. Every `dateutil` rrule carries a `threading.Lock`, so deep-copying the transition source fails with `cannot pickle '_thread.lock' object`. So the copy _shares_ the source and takes its own fresh `iter()` — which is safe precisely because `MergedIterable` is re-iterable (I verified independent/concurrent traversals stay consistent). I noted that contract in the docstring.

Yeah i believe here we understand that Iterable must always be re-iterable -- you get a new iterator every time. I had the impression that was the whole point of Iterable vs Iterator. Seems fine to note in the docstring, thank you.

allenporter (approved):
@FuNK3Y thank you for this awesome performance improvement. I didn't have the energy to support this use case, but i know its pretty popular, so thank you so much for this contribution.

I can release a new version of `ical` if you want to do the update to home assistant?

FuNK3Y:
> @FuNK3Y thank you for this awesome performance improvement. I didn't have the energy to support this use case, but i know its pretty popular, so thank you so much for this contribution.
> 
> I can release a new version of `ical` if you want to do the update to home assistant?

Thank you @allenporter for the responsiveness & guidance. That was fun.

Yes, a release would be awesome. I'll try to squeeze it in the next release of Home Assistant.

PR #656, opened 2026-07-19
fix: Add dictionary safeguards to Period before-validators

### LINKED ISSUE #655: AttributeError in Period before-validators during Union validation
In version 14.0.0, validating a model with a Union field that includes both `datetime` and `Period` (such as `list[Union[datetime.date, datetime.datetime, Period]]` in `Recurrences.rdate`) raises an unhandled `AttributeError` when Pydantic attempts to check the `Period` model.

The issue occurs because `Period`'s before-validators assume the input is always a dictionary:
1. `parse_period_fields` (in `ical/types/period.py`) calls `values.pop("value", None)`
2. `parse_parameter_values` (in `ical/types/parsing.py`) calls `values.get("params")`

When Pydantic validates a `datetime` object against the `Period` union branch, these validators receive the `datetime` object directly, causing:
`AttributeError: 'HAFakeDatetime' object has no attribute 'pop'` (or `get`).

### Proposed Fix

Add `isinstance(values, dict)` safeguards at the beginning of both functions to return early if the input is not a dictionary:

```python
# ical/types/period.py
    @model_validator(mode="before")
    @classmethod
    def parse_period_fields(cls, values: dict[str, Any]) -> dict[str, Any]:
        """Parse a rfc5545 priority value."""
        if not isinstance(values, dict):
            return values
        if not (value := values.pop("value", None)):
            return values
```

```python
# ical/types/parsing.py
def parse_parameter_values(
    cls: type[BaseModel], values: dict[str, Any]
) -> dict[str, Any]:
    """Convert property parameters to pydantic fields."""
    _LOGGER.debug("parse_parameter_values=%s", values)
    if not isinstance(values, dict):
        return values
    if params := values.get("params"):
```

allenporter:
Fixes #655

### DIFF
--- a/tests/test_event.py
+++ b/tests/test_event.py
@@ -8,6 +8,7 @@ from unittest.mock import patch
 import zoneinfo
 
 import pytest
+from freezegun import freeze_time
 from pydantic import ValidationError
 
 from ical.calendar import Calendar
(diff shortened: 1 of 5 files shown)


tests/test_event.py
```diff
@@ -651,3 +651,20 @@ def test_rfc7986_event_properties() -> None:
         in output_ics
     )
     assert "CONFERENCE;FEATURE=x-custom-feature:https://custom-conf.com" in output_ics
+
+
+def test_event_rdate_freezegun() -> None:
+    """Test parsing/validation of Event rdate with freezegun active."""
+    from freezegun import freeze_time
```
allenporter: import at top of the file

allenporter: Done. Moved the import to the top of the file.

PR #668, opened 2026-08-05
feat: add async URL fetching support via optional ical[async] extra

### LINKED ISSUE #635: Feature request: async URL fetching support (ical[async])
## Summary

No major Python iCalendar library currently supports async I/O for fetching calendars from URLs or CalDAV servers. Calendar parsing is overwhelmingly I/O-bound when fetching from remote sources, making this a natural fit for async.

## Proposed API

An `ical[async]` extras that exposes something like:

```python
from ical.calendar_stream import IcsCalendarStream

# Async URL fetching
async with IcsCalendarStream.from_url(url) as stream:
    calendar = await stream.calendar()
```

## Motivation

- **No existing libraries have this.** `icalendar`, `ics.py`, `vobject`, and `iCal-library` are all synchronous-only.
- **Real-world need:** Many applications fetch calendars from URLs (remote .ics files). Doing this synchronously blocks the event loop in async applications (FastAPI, Home Assistant, etc.).

## Implementation considerations

- Could use `aiohttp` or `httpx` as the async HTTP client (via extras dependency)
- The parsing itself can remain synchronous (it's CPU-bound and fast) — the async value is in the I/O fetch
- Consider also exposing a sync convenience wrapper (`IcsCalendarStream.from_url_sync()`) for non-async callers

proscar87:
## Problem

Fetching a remote `.ics` file today requires reading the bytes yourself and
calling the synchronous `IcsCalendarStream.from_ics()` / `calendar_from_ics()`.
Any HTTP fetch (`urllib`, `requests`, etc.) done that way blocks the event
loop, which is a real problem for asyncio-based applications — Home Assistant
is the motivating example called out in #635, since its calendar
integrations (`remote_calendar`, etc.) run entirely on asyncio and can't
afford to block on a synchronous HTTP call.

## What this adds

- `CalendarStream.from_url(url, session=None) -> CalendarStream` — async
  factory method, symmetric with the existing sync `from_ics()`. Fetches the
  ics content over HTTP asynchronously, then parses it synchronously (parsing
  is fast/CPU-bound, so there's no async value there per the issue's own
  analysis).
- `IcsCalendarStream.calendar_from_url(url, session=None) -> Calendar` —
  async convenience method for the common single-calendar case, symmetric
  with the existing sync `calendar_from_ics()`.
- Both accept an optional `session: aiohttp.ClientSession` to reuse an
  existing session/connection pool (e.g. Home Assistant's shared
  `aiohttp_client` session); otherwise a session is created and closed
  automatically for the single request.
- A new `CalendarFetchError(CalendarError)` wraps `aiohttp.ClientError` /
  non-2xx responses, consistent with how the rest of the library wraps
  failures in its own exception hierarchy.

```python
from ical.calendar_stream import IcsCalendarStream

cal = await IcsCalendarStream.calendar_from_url("https://example.com/calendar.ics")
for event in cal.timeline:
    print(event.start, event.summary)
```

### API shape vs. the issue's sketch

The issue's proposed sketch was an async context manager:

```python
async with IcsCalendarStream.from_url(url) as stream:
    calendar = await stream.calendar()
```

I intentionally did **not** implement that shape, and want to flag this
explicitly since the issue only says "something like" this API. There's no
persistent connection/resource that needs the `async with`/`stream.calendar()`
split — the fetch is a single request, and parsing is synchronous — so a
plain coroutine that returns the parsed result directly is simpler for
callers and stays symmetric with the existing `from_ics()` /
`calendar_from_ics()` naming and return types. Happy to switch to the context
manager form (or add it alongside) if you'd prefer that shape.

### Dependency choice

The issue mentions `aiohttp` or `httpx` as options. I went with `aiohttp`
since it's what your other calendar-adjacent async projects already use
(`gcal_sync`, `icaldav`), so it's the existing convention in the ecosystem
around this library and around Home Assistant. It's wired up as fully
optional:

- New `ical[async]` extra in `pyproject.toml` (`aiohttp>=3.10.0`).
- `aiohttp` is imported lazily inside `from_url()`, not at module load time,
  so `ical` without the extra installed is completely unaffected.
- If `aiohttp` isn't installed, calling `from_url()`/`calendar_from_url()`
  raises `ImportError` with a message telling the caller to
  `pip install ical[async]`.

I did not add the sync `from_url_sync()` convenience wrapper mentioned as an
"implementation consideration" in the issue — happy to add it in a follow-up
if wanted, but it felt like a separate, optional addition and I wanted to
keep this diff focused on the actual ask (async fetching).

## Testing

- `tests/test_calendar_stream.py`: 5 new tests using `pytest-aiohttp`'s
  `aiohttp_client` fixture to fake the remote server as an in-process
  `aiohttp.web.Application` on loopback — no real network calls:
  - successful fetch + parse with an owned (auto-created) session
  - successful fetch + parse reusing a caller-provided session
  - the `calendar_from_url` single-calendar convenience path
  - a non-2xx response is surfaced as `CalendarFetchError`
  - `ImportError` with the `ical[async]` install hint when `aiohttp` is
    unavailable (simulated via `monkeypatch.setitem(sys.modules, "aiohttp", None)`)
- Added `aiohttp`/`pytest-aiohttp`/`pytest-asyncio` to `requirements_dev.txt`
  as dev/test-only dependencies (same pattern already used there for `emoji`
  and `tzlocal`), and set `asyncio_mode = auto` in `pytest.ini` so these are
  the first async tests in the repo to run without a marker on every
  function.
- `uv.lock` changes are purely additive — the new `async` extra and its
  transitive deps; verified no existing pinned versions changed.

Ran the full suite the way CI does, comparing against a `git stash`'d
baseline on this same `upstream/main` checkout:

| | tests | pytest | ruff check | ruff format | ty check | codespell/yamllint |
|---|---|---|---|---|---|---|
| Before | 1428 passed, 35 skipped | pass | pass | pass | pass | pass |
| After  | 1433 passed, 35 skipped | pass | pass | pass | pass | pass |

No pre-existing failures, no regressions; the 5 new tests account for the
delta.

Also updated `README.md` with a short "Fetching a remote .ics url" quickstart
and an install note for the new extra.

Fixes #635


ical/calendar_stream.py
```diff
@@ -63,6 +82,44 @@ def from_ics(cls, content: str) -> "CalendarStream":
...
+            import aiohttp  # noqa: PLC0415
+        except ImportError as err:
+            raise ImportError(_ASYNC_EXTRA_ERROR) from err
+
+        async def _fetch(active_session: aiohttp.ClientSession) -> str:
+            try:
+                async with active_session.get(url) as response:
+                    response.raise_for_status()
+                    return await response.text()
+            except aiohttp.ClientError as err:
```
allenporter, line 111: The common scenarios here for home assistant may require another couple error error types, though we can add in a follow up: Basically: When you need to provide an auth header.

ical/calendar_stream.py
```diff
@@ -63,6 +82,44 @@ def from_ics(cls, content: str) -> "CalendarStream":
...
+                raise CalendarFetchError(
+                    f"Failed to fetch calendar from url {url!r}: {err}"
+                ) from err
+
+        if session is not None:
+            content = await _fetch(session)
+        else:
+            async with aiohttp.ClientSession() as owned_session:
+                content = await _fetch(owned_session)
+        return cls.from_ics(content)
```
allenporter, line 121: My recollection is that home assistant has to do some trickery to run this in the event loop without blocking because it may load timezones. Perhaps you can fix this in a followup PR?

allenporter (approved):
Excelent!

PR #675, opened 2026-08-12
feat(alarm): resolve VALARM triggers into absolute times

### LINKED ISSUE #651: Feature: Alarm trigger resolution API for computing absolute alarm times
## Summary

Add an API to resolve VALARM triggers into absolute `datetime` instances for a given event occurrence. This enables consumers like Home Assistant to use per-event alarm times embedded in ICS data as automation triggers, rather than requiring users to manually configure flat offsets.

### The Problem

Calendar applications store per-event reminders as VALARM components:

```
BEGIN:VEVENT
SUMMARY:Dentist Appointment
DTSTART:20250715T140000
DTEND:20250715T150000
BEGIN:VALARM
TRIGGER:-PT2H
ACTION:DISPLAY
DESCRIPTION:Time to leave for the dentist
END:VALARM
BEGIN:VALARM
TRIGGER:-PT15M
ACTION:DISPLAY
DESCRIPTION:Appointment in 15 minutes
END:VALARM
END:VEVENT
```

This event says "remind me 2 hours before AND 15 minutes before." The `ical` library already parses these into `Alarm` objects with `trigger: timedelta | datetime`. However, there is **no API to resolve these into absolute trigger times** given an event's start/end time.

### The Use Case: Home Assistant

Home Assistant's calendar trigger currently supports a fixed `offset` per automation (e.g., "trigger 15 minutes before every event on this calendar"). Users have requested the ability to honor the per-event VALARM times instead — so different events fire at different times based on their own alarm configuration.

To support this, HA needs `ical` to provide a way to ask: "given this event (or this occurrence of a recurring event), when should its alarms fire?"

### RFC 5545 Alarm Trigger Rules

The VALARM TRIGGER property has several modes (RFC 5545 §3.8.6.3):

1. **Relative to START (default):** `TRIGGER:-PT15M` → fire 15 minutes before DTSTART
2. **Relative to END:** `TRIGGER;RELATED=END:-PT5M` → fire 5 minutes before DTEND
3. **Absolute:** `TRIGGER;VALUE=DATE-TIME:20250715T120000Z` → fire at this exact time
4. **Repeated alarms:** `TRIGGER:-PT15M` + `REPEAT:3` + `DURATION:PT5M` → fire at -15m, then repeat 3 more times every 5 minutes (i.e., -15m, -10m, -5m, 0m)

### Current State

- `Event.alarm` is `list[Alarm]` — ✅ already parsed
- `Alarm.trigger` is `Union[timedelta, datetime]` — ✅ already typed
- `Alarm.repeat` and `Alarm.duration` — ✅ already modeled
- **Missing:** No RELATED parameter support (always assumes RELATED=START)
- **Missing:** No API to resolve triggers to absolute times
- **Missing:** No way to query a timeline for upcoming alarm times

---

### Option A: Method on `Alarm` — `Alarm.trigger_times()`

Add a method to `Alarm` that takes a start and end time and returns the resolved absolute trigger time(s):

```python
class Alarm(ComponentModel):
    ...

    def trigger_times(
        self,
        dtstart: datetime.datetime,
        dtend: datetime.datetime | None = None,
    ) -> list[datetime.datetime]:
        """Resolve this alarm's trigger into absolute datetime(s).

        Returns a list because REPEAT+DURATION can produce multiple times.
        The list is sorted chronologically.
        """
```

**Usage:**

```python
event = calendar.events[0]
for alarm in event.alarm:
    for trigger_time in alarm.trigger_times(event.start, event.end):
        print(f"Alarm at {trigger_time}: {alarm.description}")
```

**Pros:**
- Simple, low-level, composable
- Caller has full control over which alarms to process
- Natural home for the logic (on the Alarm itself)
- Works with both events and todos

**Cons:**
- Caller must iterate events × alarms × trigger_times manually
- Caller must handle recurring event expansion themselves
- Doesn't integrate with the Timeline

---

### Option B: Method on `Event` — `Event.alarm_trigger_times()`

Add a convenience method on `Event` that resolves all alarms at once:

```python
@dataclass
class AlarmTrigger:
    """A resolved alarm trigger with its absolute time and source alarm."""
    time: datetime.datetime
    alarm: Alarm

class Event(ComponentModel):
    ...

    def alarm_trigger_times(self) -> list[AlarmTrigger]:
        """Resolve all VALARM triggers for this event into absolute times.

        Returns a sorted list of AlarmTrigger instances. For alarms with
        REPEAT+DURATION, multiple entries are returned.
        """
```

**Usage:**

```python
event = calendar.events[0]
for trigger in event.alarm_trigger_times():
    print(f"{trigger.time}: {trigger.alarm.action} — {trigger.alarm.description}")

# Filter to upcoming alarms
now = datetime.datetime.now(tz=ZoneInfo("UTC"))
upcoming = [t for t in event.alarm_trigger_times() if t.time > now]
```

**Pros:**
- Easier for consumers — one call per event
- Returns structured data linking time back to the source alarm
- Includes all REPEAT expansions automatically

**Cons:**
- Requires `Event` to have valid start/end (which it should)
- Still requires the caller to expand recurring events first

---

### Option C: Timeline-level API — `alarm_timeline()`

Add a new timeline that yields alarm triggers across all events, integrated with recurrence expansion:

```python
@dataclass
class EventAlarmTrigger:
    """A resolved alarm trigger with its absolute time, source event, and alarm."""
    time: datetime.datetime
    event: Event
    alarm: Alarm

def alarm_timeline(events: list[Event], tzinfo: datetime.tzinfo) -> Iterable[EventAlarmTrigger]:
    """Yield alarm triggers across all events, sorted by trigger time.

    Recurring events are expanded. Each alarm on each occurrence is resolved
    to an absolute time and yielded in chronological order.
    """
```

**Usage by Home Assistant:**

```python
from ical.alarm_timeline import alarm_timeline

# Get all alarm triggers in the next 24 hours
now = datetime.datetime.now(tz=ZoneInfo("UTC"))
end = now + timedelta(days=1)

for trigger in alarm_timeline(calendar.events, local_tz):
    if trigger.time < now:
        continue
    if trigger.time > end:
        break
    schedule_notification(
        at=trigger.time,
        title=trigger.event.summary,
        body=trigger.alarm.description,
    )
```

**Pros:**
- Most powerful — handles recurrence expansion + alarm resolution in one API
- Natural extension of the existing Timeline pattern
- Exactly what HA needs: "give me all upcoming alarm times"
- Sorted chronologically across all events

**Cons:**
- Largest API surface
- More complex implementation
- May be over-engineered if consumers only need per-event resolution

---

## Recommendation

**Implement Option A + Option B (layered).** Option A provides the low-level primitive on `Alarm`, Option B provides the convenience method on `Event` that calls Option A internally. This gives consumers both fine-grained control and easy one-call usage.

Option C (alarm_timeline) is valuable but could be a follow-up built on top of A+B.

### 1. RELATED parameter support

The `Alarm` model currently does not capture the `RELATED` parameter on `TRIGGER`. This parameter can be `START` (default) or `END`, indicating whether the trigger offset is relative to DTSTART or DTEND.

The `Alarm` model should be extended:

```python
class Alarm(ComponentModel):
    trigger: Union[datetime.timedelta, datetime.datetime]
    trigger_related: str = "START"  # "START" or "END"
```

### 2. Handling edge cases

The implementation should handle:

- **Absolute triggers:** `TRIGGER;VALUE=DATE-TIME:20250715T120000Z` → return as-is
- **Relative to START:** `TRIGGER:-PT15M` → `dtstart + trigger`
- **Relative to END:** `TRIGGER;RELATED=END:-PT5M` → `dtend + trigger`
- **REPEAT + DURATION:** Generate N additional times after the initial trigger
- **All-day events:** When DTSTART is a `date`, convert to a `datetime` at midnight in the event's timezone before computing alarm time
- **Missing DTEND:** If DTEND is not specified and RELATED=END, raise or use DTSTART + DURATION

### 3. Test cases to include

- Relative trigger before start (most common case)
- Relative trigger before end
- Absolute trigger (datetime)
- Repeated alarm (TRIGGER + REPEAT + DURATION)
- Multiple alarms on a single event
- All-day event with alarm
- Event with no alarms (empty list)
- Todo with alarm (VTODO also supports VALARM)

proscar87:
Closes #651.

Implements the layered API you recommended — Option A as the primitive, Option B on top.

## `Alarm.trigger_times(dtstart, dtend=None)`

Resolves one alarm into absolute datetimes, sorted. Covers the rfc5545 modes:

- relative to start — `TRIGGER:-PT15M`
- relative to end — `TRIGGER;RELATED=END:-PT5M`
- absolute — `TRIGGER;VALUE=DATE-TIME:...`, returned unchanged and ignoring `RELATED`
- `REPEAT` + `DURATION` — one entry per firing, the initial one plus `repeat` more

It accepts `date` as well as `datetime`, so an all-day event resolves against local midnight via the existing `normalize_datetime`. A trigger relative to an end with no end raises `ValueError` rather than quietly anchoring to the start, which would move the alarm instead of failing.

## `Event.alarm_trigger_times()`

Returns `list[AlarmTrigger]`, each carrying `time` and the `alarm` it came from, sorted across all alarms rather than within each one. An alarm that cannot be resolved is skipped with a debug log instead of raising, so one malformed alarm doesn't hide the rest.

VTODO isn't given the convenience method — `Todo` importing from `event` would be circular — but `Alarm.trigger_times()` works for todos as-is, which is what the issue's todo case needs.

## RELATED turned out to be a data-losing bug, not just a gap

The issue lists "no RELATED parameter support" as missing. It's worse than missing: it was **dropped on write**. Before this change:

```
in:  TRIGGER;RELATED=END:-PT5M
out: TRIGGER:-PT5M
```

Any calendar round-tripped through this library moved every `RELATED=END` alarm from the end of its event to the start. On the example above that's a 65-minute shift, silently.

`trigger_related` is read from the parameter and written back **as a parameter**. Left to the default encoder it would have been emitted as a property of its own — `TRIGGER_RELATED:END`, not valid ics — which is why `Alarm` overrides `__encode_component__`. The default `START` is not emitted, so files that never had `RELATED` round-trip unchanged; there's a test pinning that, since emitting it would churn every file anyone writes.

## Tests

13 in `tests/test_alarm_trigger_times.py`, covering each trigger mode, the repeat expansion, the all-day case, an event with no alarms, ordering across multiple alarms, both round-trip directions, and the `ValueError`.

Full suite: **1461 passed, 35 skipped**. `ruff check`, `ruff format` and the pre-commit hooks are clean. `mypy` reports nothing in `alarm.py` or `event.py` — the 15 errors it finds elsewhere are present on an untouched `main` too.

One thing worth flagging for anyone running the suite locally: `tests/test_calendar_stream.py::test_from_url*` errors without `pytest-aiohttp` installed, on `main` as well. Not related to this change, but it cost me a few minutes.

Happy to follow up with Option C (`alarm_timeline`) on top of this if you want it.


proscar87:
Self-review turned up two defects in the first commit, both now fixed in `2071392`. Flagging them rather than quietly force-pushing, since one of them was a test that passed for the wrong reason.

**A naive absolute trigger made the result unsortable.** Every relative branch returns an aware datetime via `normalize_datetime`, but an absolute trigger was returned exactly as parsed. rfc5545 §3.8.6.3 requires absolute triggers to be UTC, so in a well-formed file this never shows — but this library reads leniently by design, and a file that omits the `Z` produces a naive value. Mixed with any relative alarm on the same event, `Event.alarm_trigger_times` raised:

```
TypeError: can't compare offset-naive and offset-aware datetimes
```

Absolute triggers now go through `normalize_datetime` too. Aware values pass through untouched, so nothing changes for conformant input.

**A negative `DURATION` produced repeats running backwards**, which contradicted the chronological ordering the docstring promises. The list is now sorted before returning.

**And I removed a branch that could not be reached.** `Event.alarm_trigger_times` caught `ValueError` from a trigger relative to a missing end — but `Event.end` always resolves to a value (falling back to the start, or start + 1 day for a date), so the exception can never occur there. Worse, the test I wrote for it admitted as much in its own docstring and then asserted the happy path. Both are gone; the `ValueError` contract still exists on `Alarm.trigger_times`, where a caller can hit it, and is still tested there.

Tests are now 15, including one per fix. Both new ones fail against the previous commit. Full suite: **1463 passed, 35 skipped**; `ruff`, `ruff format`, `ty` and the rest of the pre-commit hooks clean.

proscar87:
One more from a fuller pass, and this one was a behaviour regression rather than a gap.

**An unknown `RELATED` value failed the whole calendar.** Reading the parameter into a typed field meant pydantic validated it, so `TRIGGER;RELATED=MIDDLE:-PT5M` — previously unnoticed, because nothing read the parameter — became:

```
CalendarParseError: Failed to parse calendar ALARM component: Input should be 'START' or 'END'
```

That made the parser stricter than it was before this PR, in a library that reads leniently by design. An unrecognised value now falls back to `START`, which is exactly where the offset was already being applied when the parameter went unread. Logged at debug.

I also went looking for the cases I hadn't covered, and added tests for the two that matter most:

- **A recurring event resolves an alarm per occurrence.** This is the Home Assistant use case from the issue, and nothing exercised it. `RRULE:FREQ=DAILY;COUNT=3` through the timeline yields three distinct trigger times.
- **A VTODO alarm** through `Alarm.trigger_times` and through a full ics round-trip, since `Todo` doesn't get the convenience method.

Checked while I was in there, all clean: re-serialising all 106 ics fixtures in `tests/` produces no `RELATED=START` and no stray `TRIGGER_RELATED` property, so nothing churns; `Alarm` survives a `model_dump`/reconstruct round-trip with the field intact; and `EventStore.add` preserves it.

Now 18 tests. Each of the three fixes in this PR has a test that fails without it. Full suite **1466 passed, 35 skipped**, and `pre-commit run --all-files` is green across every hook — including `ty`, which is what caught the last one: my VTODO test passed `todo.due` where a non-optional `date` was expected.

proscar87:
Second review pass — nothing broken this time, but three gaps in coverage worth closing.

The encoder appends `RELATED` to the trigger property's `params` list. A second serialisation pass appending it again would have produced `TRIGGER;RELATED=END;RELATED=END`, and nothing held that it doesn't. It doesn't — ics output is byte-identical across repeated round trips — and there's now a test saying so.

Also pinned `REPEAT:0` firing once rather than not at all, and an `Alarm` built in Python rather than parsed, where `trigger_related` comes from the field instead of from a property this code lifted it off. That second path had no coverage.

Checked and clean without needing changes: a VALARM with two TRIGGER properties is still rejected by the existing single-value validation, and an all-day event resolves its alarm against local midnight through the full ics path.

21 tests now. Suite 1469 passed, all hooks green.

proscar87:
On Option C from #651 (`alarm_timeline`), which I offered as a follow-up — I went to design it and hit a problem worth raising before writing code, because the answer is yours rather than mine.

**Alarm order is not event order.** Measured on this branch:

| event | its alarm fires |
|---|---|
| 2025-07-10 `Cita cercana`, `TRIGGER:-PT15M` | 2025-07-10 13:45 |
| 2025-07-12 `Viaje lejano`, `TRIGGER:-P5D` | **2025-07-07** 09:00 |

The later event alarms first, by three days. So an `alarm_timeline` cannot be a loop over `timeline.included(...)` yielding as it goes — the results would come out unsorted, and the docstring in the issue promises chronological order.

**And a bounded query needs an unbounded scan.** To answer "what fires between now and tomorrow" you must look at events out to `tomorrow + longest lead time in the file`. rfc5545 puts no ceiling on that: `TRIGGER:-P1Y` is legal. So the honest options are:

1. **Explicit lookahead parameter** — `alarm_timeline(events, tzinfo, max_lead=timedelta(days=30))`, documented as "alarms with a longer lead than this are missed". Predictable cost, silently incomplete.
2. **Compute the actual maximum lead** from the parsed alarms first, then scan `window + that`. Exact, but requires a pass over every alarm before yielding anything, which gives up the laziness the rest of `Timeline` is built around.
3. **Materialise and sort** the whole range. Simplest and exact, worst for large calendars, and the least like the existing `SortableItemTimeline` design.

I lean towards 2 — the max lead is cheap to compute since alarms are already parsed, and being exact matters more here than laziness when a missed alarm is silent. But this is a shape decision in your library's core abstraction, so I'd rather ask than guess.

For context on what it unlocks: the Home Assistant case is "give me everything that fires in the next N hours" so automations can act ahead of a calendar event — pre-conditioning a room, setting a car's charge target before a trip, notifying on the event's own lead time rather than a flat offset the user configured. Option A and B already make the per-event answer possible; C is what makes it a scheduler input.

Happy to implement whichever you prefer, or to leave it if you'd rather it not exist yet.

tests/test_alarm_trigger_times.py
```diff
@@ -0,0 +1,319 @@
+"""Resolving VALARM triggers into absolute times (#651)."""
```
allenporter: Test modules names should be 1:1 with the module they are testing.  So, thing testing `ical.alarm` should be `test_alarm.py`. Thing testing `ical.event.Event` should be in `test_event.py`. etc

(AI likes to add test modules based on whatever arbitrary thing it is testing, not following the code base norms.)

ical/alarm.py
```diff
@@ -112,4 +138,102 @@ def parse_repeat_duration(self) -> Self:
...
+    def _parse_trigger_related(cls, values: Any) -> Any:
+        """Lift the RELATED parameter off TRIGGER into its own field."""
+        if not isinstance(values, dict):
+            return values
+        # Only ics parsing puts ParsedProperty here; a directly constructed
+        # Alarm passes the resolved value and has nothing to lift.
+        trigger = values.get("trigger")
+        if not isinstance(trigger, list):
+            return values
+        for prop in trigger:
```
allenporter, line 152: Does parse_parameter_values help with this? I believe we've run into this pattern a few times of needing to parse parameters on a property and assign them to a dataclass member, so there should be prior art somewhere for this that doesn't involve full manual parsing.

allenporter (changes requested):
Appreciate the great contribution!

proscar87:
Both points addressed in `49cc8d2`, though the second one I have to hand back to you rather than answer with code.

**Test module naming.** You are right, and the diagnosis is right too. `tests/test_alarm_trigger_times.py` is gone. `Alarm.trigger_times` and the `RELATED` parameter belong to `ical.alarm`, so those cases are in `tests/test_alarm.py`; `Event.alarm_trigger_times` belongs to `ical.event`, so those are in `tests/test_event.py`. All 21 cases survived the move — the two modules collect the same 76 tests they did with the extra file, so nothing was quietly dropped along the way.

**`parse_parameter_values`: I checked, and it does not reach this.** You were right that there is prior art for the pattern, but it lives one level below where I need it. `parse_parameter_values` is a `mode="before"` validator on a *value type's* model, and it reads `values["params"]` — the parameters of the single property being parsed. Six types use it that way: `CalAddress`, `Period`, `Conference`, `RelatedTo`, `Attachment`, `Image`.

`Alarm` is a `ComponentModel`, and at that level `values` maps property names to lists of `ParsedProperty`, with the parameter sitting inside one of them:

```
keys a before-validator on Alarm sees: ['action', 'description', 'trigger']
is there a 'params' key?              False
the trigger property itself:           ParsedProperty(name='trigger', value='-PT5M',
                                         params=[ParsedPropertyParameter(name='RELATED', values=['END'])])
```

Calling `parse_parameter_values(Alarm, values)` on that returns the input unchanged. I could not find any component in the tree that lifts a property parameter into a component field — `get_parameter_value` appears only in `types/data_types.py` (for `VALUE`) and in `parsing/property.py` itself. So there is no component-level precedent to follow, which is consistent with your instinct that the precedent exists, just not at this altitude.

**The idiomatic version of this would change public API, so it is your call.** The shape that matches the prior art is `RelatedTo`: a registered data type carrying the value plus its parameter, with `__parse_property_value__`, `parse_parameter_values` and `__encode_property__` doing the work, and no manual lifting or `__encode_component__` override on my side. Applying it to `TRIGGER` means the trigger becomes that type, so `Alarm.trigger` stops being `datetime.timedelta | datetime.datetime`. That is a breaking change to a documented field, and I would rather you decide it than discover it. I am happy to write it if you want it; if you would rather keep `Alarm.trigger` as it is, the manual lift stays as the price of not breaking it.

Still open from my side, and yours to answer: the `alarm_timeline` design question above about alarm order not being event order.

ical/event.py
```diff
@@ -339,6 +351,22 @@ def computed_duration(self) -> datetime.timedelta:
             return self.duration
         return self.end - self.start
 
+    def alarm_trigger_times(self) -> list[AlarmTrigger]:
```
allenporter, line 354: We need to caveat that this does not handle recurring events.

Can we convince ourselves that this API will still make sense when we handle recurring events?

allenporter (changes requested):
Please reply / resolve comments directly as it is hard to follow the conversation back and forth when coming back to review this. Thanks for the contribution.

tests/test_alarm_trigger_times.py
proscar87: Done in 49cc8d2f, pushed on Aug 17 — sorry for not saying so here at the time. `test_alarm_trigger_times.py` is gone: the `Alarm.trigger_times` cases moved into `tests/test_alarm.py` and the `Event.alarm_trigger_times` ones into `tests/test_event.py`.

ical/alarm.py
proscar87: I looked at it, and I do not think it reaches this case — but the alternative that would let it is a public API decision, so it is yours to make.

`parse_parameter_values` reads `values["params"]` and maps a property's own parameters onto fields of the model being validated. That fits every current user — `CalAddress`, `Period`, `Relation`, `Attachment`, `Image`, `Conference` — because each of those models *is* one property's value, so the parameters and the fields belong to the same object. `Alarm` is a component, and `RELATED` sits on its child `TRIGGER` property, so there is no `params` key at the `Alarm` level for the helper to find.

The prior art for reading a parameter off a child property inside a component looks manual, done at the point where the parameter matters: `calendar.py:200` and `types/date_time.py:33` for `TZID`, `types/data_types.py:150` for `VALUE`.

The shape that would use the helper is making the trigger a value type that carries its own `RELATED`, so `alarm.trigger` becomes an object and `trigger_related` goes away. That is a change to the public field shape rather than a refactor, so I left it alone — happy to do it that way if you would rather have it.

ical/event.py
proscar87: Caveat added in ecf29812.

On whether the API survives recurrence: I think it does, because it resolves against `self.start` and `self.end` rather than against the master event. `RecurAdapter.build` hands out `model_copy(update={"dtstart", "dtend", "recurrence_id"})`, which carries the same `alarm` list, so calling this on a timeline occurrence answers for that occurrence. Checked it rather than assuming — `FREQ=DAILY;COUNT=3` with `TRIGGER:-PT15M` off a 09:00Z event gives 08:45Z on each of the three days through the timeline, while the master alone answers only for its own instance. `test_alarm_trigger_times_for_each_occurrence` in `tests/test_event.py` pins that.

So no signature change should be needed when recurrence support lands. What could be added later is a helper that walks the timeline for you and returns the trigger times over a window, but that reads like a separate API rather than a change to this one.

Unrelated, while you are here: `ty` is red on this branch at `recur_adapter.py:84`. It fails identically at the branch's merge base and passes on current `main`, so it wants a rebase and not a code change. I held off rebasing so these threads stay readable — say the word and I will.

Issue #70, opened 2022-08-10
Recurrence date/time and timezone validation rules


allenporter:
**Is your feature request related to a problem? Please describe.**
There are explicit rules in https://www.rfc-editor.org/rfc/rfc5545#section-3.3.10 that describe the validations that need to happen for recurrence dates and timezones which are not implemented. These are across objects, so a little more complicated to validate, but can likely happen in the parent component.

- Recurrence timezone formattings
- Recurrence datetime field encoded properly

allenporter:
Tests are needed to ensure that recurring all day events are expanded with the right timezone rules

allenporter:
### Proposed Implementation Plan

**Goal**: Implement validation rules specified in RFC 5545 section 3.3.10 for recurrence dates and timezones (e.g., verifying daylight saving time matching between start and until values).

**Proposed Changes**:
Implement validation logic inside `validate_until_dtstart` in `ical/component.py`:
* Verify that both `DTSTART` and `UNTIL` are either floating time or both are UTC/timezone-localized.
* Verify that if `DTSTART` is `DATE`, `UNTIL` must also be `DATE`.
* Raise a validation error if these rules are violated and compatibility mode is not active.

PR #676, opened 2026-08-12
fix(recur): reject a naive UNTIL against a timezone-aware DTSTART

proscar87:
Part of #70. Two related holes in how a naive DATE-TIME is handled against a timezone-aware `DTSTART`.

## Example

Two calendars, both accepted by the parser today:

```
DTSTART:20250715T140000Z
RRULE:FREQ=DAILY;UNTIL=20250718T140000        <- no Z
```

```
DTSTART:20250715T140000Z
RRULE:FREQ=DAILY;COUNT=3
EXDATE:20250716T140000                        <- no Z
```

## Old behavior

Both parse cleanly, and then fail when something expands them:

| | on parse | on expansion |
|---|---|---|
| naive `UNTIL` | accepted | `ValueError: RRULE UNTIL values must be specified in UTC when DTSTART is timezone-aware` |
| naive `EXDATE` / `RDATE` | accepted | `RecurrenceError: Error evaluating recurrence rule (...): can't compare offset-naive and offset-aware datetimes`, from `ical/iter.py:217` wrapping a `TypeError` out of `dateutil` |

Neither message names the event or the rule, and both surface while a consumer is rendering a calendar rather than while loading it. For Home Assistant that is a broken calendar card instead of a config entry that fails with a reason. In the `EXDATE` case nothing is excluded at all — the whole expansion dies rather than the date being applied.

## Expected / new behavior

- A naive `UNTIL` against an aware `DTSTART` is rejected during validation, alongside the other rfc5545 §3.3.10 rules: `DTSTART was timezone-aware but UNTIL had no timezone: UNTIL must be specified in UTC`.
- Under `dtstart_until_compat` it is read as UTC instead — a producer that omits the `Z` means UTC — mirroring how the DATE/DATE-TIME mismatch a dozen lines above is already handled.
- A naive `EXDATE`/`RDATE` against an aware `DTSTART` is read as `DTSTART`'s zone, the wall time the producer wrote, so the date applies instead of the expansion failing.

## Caveats, risks, rollout

- **Outside compat mode this turns a late failure into an early one, not a failure into a success.** A calendar with a naive `UNTIL` used to load and break at render; now it fails at parse with a message that says why. That is the intent, but it is a behavior change for any consumer that was catching the render-time error.
- **The one judgement call is the flag.** I reused the existing Google-scoped `dtstart_until_compat` rather than adding one, since it covers exactly this `DTSTART`/`UNTIL` relationship in the same RFC section. Happy to split it into its own flag, or to make the coercion unconditional — unconditional would turn every affected calendar from "fails somewhere" into "works", at the cost of assuming UTC for a producer that might have meant local time.
- **The mirror cases are left alone on purpose.** An aware recurrence date against a floating `DTSTART` keeps its offset, because dropping it would silently rewrite the value on the way back out; a floating `DTSTART` with a floating `UNTIL` stays valid. Both are pinned by tests.
- **The `EXDATE`/`RDATE` half assumes wall time, not instant.** A producer that wrote a naive value meaning UTC, while `DTSTART` is in another zone, is now aligned to that zone. The file gives no way to tell the two apart, and `_as_datetime` already makes this same assumption for a DATE value a few lines above.

## Checks

14 tests, of which **8 fail without the change**: three `UNTIL` rejections, two compat coercions, three `EXDATE`/`RDATE` alignments. The other six pin what must not change.

Rebased onto `main` at e4df982, where all `pre-commit` hooks pass, `ty` included. Full suite there: 1468 passed, 35 skipped, and 20 failures in `tests/parsing/test_property.py` that reproduce identically on clean `main` with none of these commits applied — snapshot mismatches from my local environment rather than anything in this change, since CI is green on that commit.

<details>
<summary>How this was diagnosed</summary>

The item on #70 was "verify daylight saving time in until and start is the same", which read like a missing assertion. `_adjust_recurrence_date` already enforced most of §3.3.10; its timezone branch was:

```python
if date_value.utcoffset():
    raise ValueError("DTSTART had UTC or local and UNTIL must be UTC")
```

`utcoffset()` is falsy for UTC, which is correct and intended — but it is also `None` for a naive value, so `DTSTART:20250715T140000Z` with `UNTIL=20250718T140000` passed validation, as did the same thing with a `TZID` reference. The hole was the guard's shape, not a missing rule.

Chasing where that surfaced led to the second hole. `_as_datetime` gives a DATE value `dtstart`'s `tzinfo` a few lines up, but passed a naive DATE-TIME through untouched, so a naive `EXDATE` reached `dateutil` and the comparison blew up inside the iterator. That is why this PR touches two functions instead of one.

Every `.ics` fixture under `tests/` was re-parsed before and after: 19 fail either way, none newly. Several of those are invalid on purpose.

</details>


proscar87:
Second review pass. One real gap: `validate_until_dtstart` is shared by VEVENT, VTODO and VJOURNAL, and I had only exercised events.

I confirmed the fix behaves identically on all three — a VTODO and a VJOURNAL with a naive UNTIL against an aware DTSTART are both rejected, and compat coerces both. Added VTODO tests; **neither VTODO nor VJOURNAL had any UNTIL coverage at all** before this, which is probably why the gap in `utcoffset()` survived so long.

8 tests now, 5 of which fail without the change. Suite 1456 passed, all hooks green.

proscar87:
Added the second checkbox from #70 — "recurrence datetime field encoded properly" — since it turned out to be the same defect as the UNTIL one, in the same function's neighbourhood.

`_as_datetime` gives a DATE value `dtstart`'s tzinfo when converting it, but a DATE-TIME that disagreed about awareness was returned untouched. The recurrence iterator then can't compare it against `dtstart`, so **the whole expansion raises rather than the date being ignored** — the event stops producing occurrences entirely.

Measured against `DTSTART:20250715T140000Z` with `RRULE:FREQ=DAILY;COUNT=3`:

| | before | after |
|---|---|---|
| `EXDATE:20250716T140000` | `RecurrenceError` | 15th, 17th |
| `RDATE:20250720T140000` | `RecurrenceError` | 15th, 16th, 17th, 20th |
| `EXDATE:20250716T140000Z` | 15th, 17th | unchanged |
| `EXDATE;TZID=America/Denver:20250716T080000` | 15th, 17th | unchanged |

Aligning a naive value to `dtstart`'s zone is what the DATE branch four lines above already assumes, so this isn't a new policy — it's extending an existing one to the case it skipped. The mirror (aware value, naive `dtstart`) and the different-named-zone case both already worked and are now pinned so they stay that way.

14 tests in the file, 6 for this. Three fail without the change; the three that pin unchanged behaviour pass either way, which is what they're for.

Full suite **1462 passed, 35 skipped**; `pre-commit run --all-files` green.

**A note on my own verification, since it nearly fooled me.** My first negative check cut the wrong lines out of `_as_datetime` — `"    return date_value"` is a substring of `"        return date_value.replace(...)"`, so the edit left the alignment applied unconditionally instead of removing it. That made a *different* test fail and looked, at a glance, like the tests catching the change. Redone properly, the three that should fail do.

That leaves one unticked box on #70: "recurrence timezone formattings". I haven't found a failing case for it yet and would rather not pad this PR with a change I can't demonstrate is needed.

tests/test_until_timezone_rules.py
```diff
@@ -0,0 +1,228 @@
+"""A naive UNTIL against a timezone-aware DTSTART (#70).
```
allenporter: Please add these tests to a test module name of the module under test. (e.g. test_component.py and/or wherever calendar_stream is tested, or next to the test compat cases for compat mode fixes.

tests/test_until_timezone_rules.py
```diff
@@ -0,0 +1,228 @@
...
+UID:1
+{dtstart}
+RRULE:FREQ=DAILY;UNTIL={until}
+SUMMARY:Standup
+END:VEVENT
+END:VCALENDAR
+"""
+
+
+def _until(calendar) -> datetime.datetime | datetime.date:
```
allenporter: Can you type the calendar argument please

allenporter (changes requested):
Thanks for making these improvements.  May take me a few passes to review/process since datetime bugs always take awhile for me to internalize.

proscar87:
Both points addressed in `174c077` — and doing the first one turned up a defect in my own second commit, so thank you for the nudge.

**Test module naming.** Both fixes live in `ical/component.py`, so the parsing and expansion cases moved to `tests/test_component.py`, and the compat-mode cases joined the other DTSTART/UNTIL compat cases in `tests/compat/test_dtstart_until_compat.py`. `tests/test_until_timezone_rules.py` is gone. The helper that reads a calendar is typed.

**The defect: `_as_datetime` was aligning a direction that was never broken.** I wrote the condition as "the two disagree about awareness", but only one of those disagreements fails. A naive `EXDATE`/`RDATE` against an aware `DTSTART` is the broken one — the recurrence iterator cannot compare them, and the whole expansion raises. The mirror case, an aware value against a floating `DTSTART`, already expanded correctly. Aligning it too dropped the offset:

```
input:                DTSTART:20250715T140000
                      EXDATE;TZID=America/Denver:20250716T140000
                      RDATE:20250725T190000Z

before (my commit):   EXDATE:20250716T140000          ← zone gone
                      RDATE:20250725T190000           ← Z gone
after:                EXDATE;TZID=America/Denver:20250716T140000
                      RDATE:20250725T190000Z
```

A different instant than the producer wrote, on a plain read-then-write round trip. My test for that case asserted only the number of occurrences, which is identical either way, so it passed while the value underneath it changed — the same shape of mistake as the one I flagged earlier in this PR. The condition is narrowed to the direction that was actually broken, and there is now a test asserting the round trip byte-for-byte on both properties.

Differential, so the tests are not just decorative: 8 of them fail against `main`, and the new round-trip one fails against the version of this PR that you reviewed.

`ruff check`, `ruff format --check`, `ty check` and the full suite (1462 passed) are clean.

allenporter:
Hello, trying to catch back up on this issue, but there is a lot of information and a long journey.

Could we step back and simplify the PR to describe:
- an example case
- old behavior
- expected / new behavior
plus: any other caveats / risks or update/rollout considerations

It would be easier than following the step by step journey taken diagnosing this. We can stick that in the appending/details at the bottom.

PR #682, opened 2026-08-19
feat(diff): compare two versions of the same calendar

### LINKED ISSUE #677: Feature: compare two versions of the same calendar (what changed between fetches)
## Feature request: comparing two versions of the same calendar

A gap I ran into that isn't on the roadmap, and which follows directly from `ical[async]` (#668): once you can fetch a calendar from a URL, the next thing you want is to fetch it *again* and know what moved.

### The case

A subscribed calendar — a sports fixture list, a class timetable, a shared team calendar. It is re-fetched on a timer. Between two fetches, an event's time changes:

```
SUMMARY:Race            SUMMARY:Race
DTSTART:20251026T200000Z    →   DTSTART:20251026T203000Z
SEQUENCE:0              SEQUENCE:1
```

Today the library gives you two `Calendar` objects and nothing to compare them with. Every consumer writes the same loop: index the old one by `uid`, walk the new one, compare fields by hand. I just wrote it to check, and it is easy to write *wrongly* — the naive version keys on `uid` alone, which is incorrect the moment a recurring event has a modified instance, since those share a `uid` and differ only by `recurrence_id`.

### Why the library is the right place

rfc5545 already defines the semantics, and `ical` already parses all of it:

- `SEQUENCE` (`event.py:269`) is the spec's own revision counter, monotonically incremented on each significant change
- `LAST-MODIFIED` (`event.py:184`)
- `uid` + `recurrence_id` is the actual identity of an occurrence

So the pieces are present and nothing uses them to answer the question. A consumer reimplementing this gets the identity rule wrong before they get the comparison wrong.

### Rough shape

```python
@dataclass
class CalendarDiff:
    added: list[Event]
    removed: list[Event]
    modified: list[tuple[Event, Event]]   # (before, after)

def diff_calendars(before: Calendar, after: Calendar) -> CalendarDiff: ...
```

Open questions I would want your answer on before writing any of it:

1. **What counts as modified?** `SEQUENCE` increasing is the spec's answer, but plenty of producers never increment it. Falling back to field comparison is more useful and less principled. Both? Configurable?
2. **Which fields?** Comparing everything reports noise like `DTSTAMP` changing on every export. A curated set (`dtstart`, `dtend`, `summary`, `location`, `status`) is more useful and more opinionated.
3. **Recurring events** — diff the rule, or the expanded occurrences within a window? The first is cheap and exact about intent; the second is what a consumer scheduling actions actually needs.
4. **Does this belong here at all**, or is it consumer territory? I think the identity rule alone justifies it, but it is your library.

Happy to implement it if the shape is welcome — and equally happy to be told it is out of scope. I would rather ask than arrive with a design you did not want.

proscar87:
Fixes #677.

## Summary

`diff_events(old, new)` compares the events of two fetches of the same
calendar and reports `added`, `removed`, and `changed` (old and new value
side by side), matching occurrences by `(uid, recurrence_id)` -- the identity
rfc5545 defines for a specific instance of a possibly-recurring event, not by
`uid` alone. `uid` alone conflates a whole series with a single modified
instance within it, since both carry the same `uid` and differ only by
`recurrence_id`.

## A design decision that needed checking before I could trust it: `dtstamp`

Naive full-field equality would compare `dtstamp` along with everything else.
I checked what that field actually means before relying on it: it records
when the ics object was generated, not the event's own content, and some
producers regenerate it on every export even when nothing about the event
changed. Measured directly -- two parses of the same event differing only in
`DTSTAMP` compare unequal under plain `==`:

```
solo DTSTAMP distinto -> e1 == e2: False
con dtstamp neutralizado -> igual:  True
```

Comparing it as-is would report every event from such a producer as changed
on every single fetch, which defeats the entire point of a diff. It's
excluded from the content comparison; every other field, including `extras`
(passthrough unknown properties), is compared as-is.

## Every design decision is pinned by a test that fails if reverted

Checked by mutating the implementation back to the wrong version and
confirming the *right* test breaks each time, not just that something
breaks:

- Removing the `dtstamp` exclusion → `test_regenerated_dtstamp_alone_is_not_a_change`
  fails (and so does the plain unchanged-event case).
- Matching by `uid` alone instead of `(uid, recurrence_id)` →
  `test_a_modified_recurrence_instance_is_matched_by_recurrence_id_not_uid_alone`
  fails.
- Comparing `new` against `old` without deduplicating `new` the same way
  `old` is deduplicated → `test_a_duplicate_identity_within_one_list_keeps_the_last_occurrence`
  fails. This one caught a real bug in my own first draft: I built a dedup
  dict for `old` but iterated `new` as a raw list, so a duplicate identity in
  `new` would be compared and reported twice instead of once. Caught by the
  test before this was ever run against the good version once.

## Two things worth flagging, found on a second pass

**A narrow, non-blocking equality gap.** `RecurrenceId` bolts `tzinfo` and
`range` onto a `str` subclass without a custom `__eq__` -- confirmed by
constructing two instances with the same string and different tzinfo/range
and comparing them: `True`. Two occurrences with the same recurrence_id
string but different tzinfo/range would compare equal on that field.
Harmless in practice: recurrence_id is the identity key `diff_events` matches
on, so it's always equal-by-construction within any pair that gets compared.
Checked it isn't a wider pattern first -- `Priority` and `Uri` are the only
other `str`/`int` subclasses in the type system, and neither bolts on extra
state. Documented in `_identity`'s docstring (`88c8144a`).

**An API scope question I should raise rather than decide silently.** The
issue's title says "compare two versions of the same calendar," but this PR
takes `list[Event]`, not `Calendar`. I scoped it that way because the
issue's own example is a bare `VEVENT`, but a `diff_calendars(old: Calendar,
new: Calendar) -> EventDiff` convenience wrapper would be more discoverable
and closer to the title. Happy to add it if you'd rather have it than the
current signature.

## Scope

`Event` only, matching the issue's example. `Todo` and `Journal` carry the
identical `uid` / `recurrence_id` / `sequence` / `last_modified` shape, so
the same approach would extend to them with no new design work -- flagging
that as a natural follow-up rather than building it here unrequested.

## Validation

- `pytest`: 1484 passed, 35 skipped (full suite, this branch).
- `ruff check` / `ruff format --check`: clean.
- `ty check`: clean.
- `pre-commit run --all-files`: clean (trailing-whitespace, ruff, ty, codespell).
- Confirmed `ical.diff` does not exist on `main` (`git stash` + import check),
  so this is new capability, not a fix to something already there.

### DIFF
--- a/ical/__init__.py
+++ b/ical/__init__.py
@@ -6,6 +6,7 @@ __all__ = [
     "alarm",
     "calendar",
     "calendar_stream",
+    "diff",
     "event",
     "freebusy",
     "journal",
--- a/ical/diff.py
+++ b/ical/diff.py
@@ -24,6 +24,15 @@ def _identity(event: Event) -> tuple[str, str]:
     specific instance within it additionally needs `recurrence_id`. Two
     events sharing both identify the same occurrence, even across separate
     parses of separate fetches.
+
+    Compared as a plain string. `RecurrenceId` also carries `tzinfo` and
+    `range`, parsed from the TZID/RANGE property parameters, but they are
+    ordinary instance attributes bolted onto a `str` subclass rather than
+    part of its `__eq__` -- so two occurrences with the same recurrence_id
+    string but different tzinfo/range would still identify the same
+    occurrence here. That can't produce a missed content change: since
+    recurrence_id is the identity key, it is always equal-by-construction
+    within a pair `_content_equal` ever compares.
     """
     return (event.uid, event.recurrence_id or "")
 
--- a/ical/diff.py
+++ b/ical/diff.py
@@ -1,0 +1,108 @@
+"""Comparing two versions of the same set of events.
+
+A calendar fetched on a recurring basis -- a sports fixture list, a class
+timetable, a shared team calendar -- changes between fetches. This module
+answers what changed, using the identity rule rfc5545 already defines
+rather than a naive comparison by `uid` alone, which is wrong the moment a
+recurring event has a modified instance: those share a `uid` with the
+series and differ only by `recurrence_id`.
+"""
+
+from __future__ import annotations
+
+import dataclasses
+
+from .event import Event
+
+__all__ = ["EventChange", "EventDiff", "diff_events"]
+
+
+def _identity(event: Event) -> tuple[str, str]:
+    """Return the rfc5545 identity of an event occurrence.
+
+    The full range of a recurrence set is referenced by `uid` alone; a
+    specific instance within it additionally needs `recurrence_id`. Two
+    events sharing both identify the same occurrence, even across separate
+    parses of separate fetches.
+    """
+    return (event.uid, event.recurrence_id or "")
+
+
+def _content_equal(old: Event, new: Event) -> bool:
+    """Return whether two same-identity events have the same content.
+
+    `dtstamp` is excluded: it records when the ics object was generated,
+    not the event's own content, and some producers regenerate it on every
+    export of an otherwise unchanged event. Comparing it would report every
+    event from such a producer as changed, defeating the point of a diff.
+    Every other field, including `extras` (passthrough unknown properties),
+    is compared -- a producer that encodes something volatile there is a
+    real gap, not one to guess a fix for without a sample to test against.
+    """
+    return old.model_copy(update={"dtstamp": None}) == new.model_copy(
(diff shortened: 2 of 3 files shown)


allenporter:
Marking as draft for now while we continue discussion in https://github.com/allenporter/ical/issues/677#issuecomment-5386969257

